PR #11, opened 2017-11-30
Migrate to Lodash Modules

lukeed:
Importing modules from Lodash via `lodash/*` is rarely a clean import. You need to configure babel/webpack heavily and/or use the [babel plugin](https://www.npmjs.com/package/babel-plugin-lodash), which typically still requires extra config.

Any users who wire up this great lib for Module imports will _also_ need to be aware of the full `lodash` import.

It's far better to just use the modular `lodash.*` packages instead. John David Dalton recommends this too! It's more declarative, and is the most reliable way to avoid needless lodash imports.

For example, these changes alone dropped the 95kB minified file to 61kB! 🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "component-type": "^1.2.1",
-    "lodash": "^4.17.4"
+    "lodash.clonedeep": "^4.5.0"
   },
   "devDependencies": {
     "babel-cli": "^6.10.1",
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
@lukeed ah, I didn't realize that! I thought that the `/` imports were the most recently recommended way. But this way actually leads to smaller builds huh? Thank you for this.

Also, I think there are probably other super-low-hanging fruit for reducing the file size that I haven't checked out yet, if you find others that would be amazing! Thank you!

ianstormtaylor:
Also, just added the a GZIP'd size badge to the Readme, to keep us honest!

lukeed:
No problem! I haven't looked through the actual code yet, but I plan to soon. I'd be happy to continue with any reductions 😄 

I had a similar library for a bit, but abandoned the parent project... This looks much nicer anyway.

lukeed:
Actually, I have my next target 😇

PR #14, opened 2017-11-30
Rework Build tools

### LINKED ISSUE #12: Build script fails on Windows
```
λ yarn run build
yarn run v1.3.2
$ yarn run build:lib && yarn run build:max && yarn run build:min
$ babel ./src --out-dir ./lib
src\default-types.js -> lib\default-types.js
src\index.js -> lib\index.js
src\struct-error.js -> lib\struct-error.js
src\superstruct.js -> lib\superstruct.js
$ mkdir -p ./dist && browserify ./src/index.js --transform babelify --standalone Superstruct > ./dist/superstruct.js
The syntax of the command is incorrect.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'll try to work on this.

lukeed:
***Why?***

Rollup is known for consistently producing the smallest bundles. In this case, the biggest reason is because Browserify's overhead is much bigger than Rollup. Rollup also inlines its imports directly within the bundles, whereas Browserify leaves `require` statements, each of which needs its own interop wrapper. 

...That's also to say that the Browserify build has to _deal with_ require statements at runtime, which seems like a silly step to take since we're pre-compiling anyway.

Bublé also consistently produces smaller ES5 code _and_ with less configuration. It exposes its own `buble/register` which is now in the tests. Because of this, the tests' elapsed time has dropped from `~750ms` to `~115ms`!

screen shot 2017-11-29 at 8 08 52 pm

***Commands***

As for the actual `"build"` command, I imported a `builder.js` file I use across a bunch of my modules. It produces 3 files in 3 different formats:

* **ES Module**: `lib/superstruct.es.js`
* **CommonJS**: `lib/superstruct.js`
* **UMD**: `lib/superstruct.min.js`

I've updated `package.json`'s entries to expose these three correctly. 

The main benefit here is that Webpack/Rollup users get to use partial `import`s for free, and Node.js servers can use the CJS module (sans `require`s!) as normal. 

The `"watch"` command was updated to use Rollup directly. Nothing extra is needed. I have it producing the CJS/main file only. Don't think the others are needed here.

***Results***

The final GZIP size of the UMD file is `5.11 kB`! 🎉 Of course, this number is actually smaller if you compress the other two files (no UMD header).

***Suggestions***

* It's a personal thing, I know, but the `"clean"` command seems useless to me. 😜 

* I'd also remove `.npmignore` in favor of specifying a `files` key within `package.json`. I personally only allow the `lib` directory -- imo, devs can lookup the repo for docs & examples. But you can definitely include `docs` and `examples` here too if you want. 😄 

---

Closes #12 because Rollup handles this internally.

---

Sorry if this was more of an overhaul than you would have liked @ianstormtaylor 😅


ianstormtaylor:
Hey @lukeed thank you for this! This is super interesting. I haven't had a chance to check out Rollup yet, so I appreciate the detailed write up. I'd definitely be down to switch to using Rollup if it does save a bunch more on size.

I've just done a partial refactor of the build process on another branch, and merged it into master. This fixed a few different low-hanging fruit things like minifying, removing some unnecessary stuff, etc. It looks like I've got the `./dist/superstruct.min.js` file down to `4.17kb` GZIP'd. 

Would you be able to rebase/merge from the existing master and see how much Rollup will save us on top of that? I'm just curious to get a good comparison, since some of the stuff was just me not having set things up right.

Thank you!

lukeed:
Hey, will do~! 

Yeah, the broken minifying was an easy fix as it was, but it made most sense to me to fix the exports too while I was at it.

lukeed:
Very cool. I was going to suggest `kind-of` and `clone-deep` too

lukeed:
screen shot 2017-11-30 at 5 52 24 pm

👆 _Still the UMD bundle_

sudo-suhas:
@lukeed How about adding the `engines` field to package.json to indicate supported node version?

lukeed:
Hey @sudo-suhas, there isn't a reason to really. The bundles are compiled to ES5 which runs everywhere. Only the build script requires `6.x` but the engines key doesn't reflect that value.

sudo-suhas:
It might not be immediately obvious if the package can be used in node 4. Though I am not sure how many people still develop on node 4.

ianstormtaylor:
Okay, @lukeed thank you! I'm definitely convinced.

As you guessed, this PR is a little bit too much for me in terms of changes to the build setup. But I've just pulled a variant of the Rollup configuration you've got here, so the newest release is built with Rollup and exposes the `package.module` for ES imports too.

Thank you!

(I'm going to close this since we've got the benefits.)

lukeed:
Ah, bummer. Would have been happy to accommodate. 

FWIW, you're still losing out the benefits of Bublé -- I would guess 0.5kb in savings. 

And now that you're using `rollup`, remove the `mkdir` calls. That alone will close #12 because Rollup creates directories as needed.

ianstormtaylor:
Ah sorry, I didn't realize that. I just figured that me asking you for lots of tiny little fixes would get annoying and feel like I was foisting work onto you, so I figured it would be nicer to just do the switches myself instead of pushing it onto you.

I hear you on Bublé, I think I might still make that change to get those extra savings. I was originally worried about using a less-standardized one, but after going through the codebase it seems like it won't be a big deal since it's pretty tiny to begin with, and that definitely seems worth the `0.5kb`. Thanks!

I just realized the reason for the `mkdir` failures I was seeing was because I was piping to `uglify` so the `./umd` directory wasn't being created automatically. Good call, I'll fix that.

ianstormtaylor:
Hey @lukeed have you ever handled removing dev-only code paths with Rollup/Bublé? Asked about it here: https://github.com/rollup/rollup/issues/1770 but it seems not supported right now. That's one of the things I was using Babel for in the latest commits, to allow `invariant` with long and helpful error strings that get removed in the UMD builds.

lukeed:
No worries! It's your library anyway, so totally your call on all matters :D

For dev-only code paths, you can take a cue from the frontend world. You can wrap your invariant calls (or any dev code) with:

```js
if (process.env.NODE_ENV === 'development') {
  invariant (...)
}
```

Then if you use `rollup-plugin-replace` to update that `process.env.NODE_ENV` value.

On production builds, that code block will be `if (false)` so Uglify will remove it because dead code.

ianstormtaylor:
Yeah true I guess I can just ditch the invariant package anyways and throw errors with the `if` statements. Was just hoping for a super terse solution haha.

lukeed:
You can always create your own little utility wrapper for dynamic `invariant()` calls based on `NODE_ENV` value. That seems to be all that the `dev-expression` plugin is doing.

```js
if (process.env.NODE_ENV !== 'production') {
  // verbose dev errors, via invariant
  // auto-removed for production builds
} else {
  // simpler invariant() call here
  // kept in prod
}
```

PR #30, opened 2017-12-02
Fix typo

chocolateboy:
s/Javascript/JavaScript/g


ianstormtaylor:
Fair enough 😄  thanks @chocolateboy!

I really wish it wasn't spelled that way haha, but no point in fighting it.

PR #31, opened 2017-12-03
fix throwing errors example with expected values

RayBenefield:
The value passed in in the example is actually `true` not `false`. So just a quick correction of the example.


ianstormtaylor:
Ah good catch, thanks @RayBenefield!

PR #49, opened 2017-12-09
Fix example

jlcarvalho:
Import `struct` on custom data types example.


citycide:
This isn't necessary. You're creating your own `struct` using the `superstruct` factory:

```js
import { superstruct } from 'superstruct'

// create it
const struct = superstruct({
  types: { /*...*/ }
})

// then use it
const User = struct({ /*...*/ })
```

jlcarvalho:
I didn't noticed. You're absolutely right. Sorry. ;)

ianstormtaylor:
@jlcarvalho no worries!

Issue #58, opened 2017-12-17
Cannot convert object to primitive value


istarkov:
This line https://github.com/ianstormtaylor/superstruct/blob/9dca528783f9db1aaf0b6bb5e94d34c291f120cb/src/error.js#L12 caused an error `Cannot convert object to primitive value`.

Modern js frameworks like to implement objects without prototype like:
 
```javascript
      const a = Object.create(null);
      a.x = 1;
      // The line below will fail with "Cannot convert object to primitive value"
      const www = `jhjh ${a}`;
```

So every time I met an error in such object instead of superstruct error I get  `Cannot convert object to primitive value`.

I'll try to provide a fix

PR #59, opened 2017-12-17
Cannot convert object to primitive value

istarkov:
Add failing test #58


istarkov:
I'm not sure about is custom toString is needed in production, but it's needed at dev.
For now I use `superstruct` for additional validation of `graphql` input, 
having that it produces all the input objects via `Object.create` validate method throws.

istarkov:
Also to make error message more readable somehow, to show instead of `[Object object]` seems it will be better to use `JSON.stringify` instead of custom `toString`, having that custom `toString` have no protection against circular dependencies.

test/index.js
```diff
@@ -29,6 +29,10 @@ describe('superstruct', () => {
           }
 
           else if ('error' in module) {
+            const [ err ] = Struct.validate(data)
+            const act = pick(err, 'type', 'path', 'value')
+            assert.deepEqual(act, module.error)
+
```
ianstormtaylor: Hey @istarkov what's the need for this extra logic? Is it just to ensure that `.validate()` works too? I think we should only have one of these here, and then have the `validate/test/assert` methods tested elsewhere.

I'd be open to only testing with `validate` here so that we dont have to use `assert.throws` if that's what you were after.

ianstormtaylor:
Awesome, thanks @istarkov!

Issue #8, opened 2017-11-29
consider adding value coercion


ianstormtaylor:
This can sometimes be useful, for example when you require numbers, but a string of a number was passed in. If there was a coercion step before validation it would account for that.

mycaule:
You can already do this.

```
imageHeight: 'string | number',
```

mycaule:
Here is a sample piece of code for an implementation of the [DuckDuckGo Instant Answer API](https://duckduckgo.com/api).

https://github.com/mycaule/knowledge-graph-js/blob/master/services/duckduckgo.js

```
  ImageIsLogo: 'empty | number',
  ImageHeight: 'empty | number',
  ImageWidth: 'empty | number',
```

 The response can be pretty complex. I retro-engineered their contract with your library.

huanz:
I'm also looking forward to this and for `struct.enum` ，eg：

```javascript
{
    status: shape.enum([0, 1, 2])
}
```

the params passed from a request may be a string，now i must do like this

```javascript
{
    status: shape.enum([0, 1, 2, '0', '1', '2'])
}
```
I hope there's a better way to deal with it

ianstormtaylor:
The main place this has come up for me has been parsing query parameters. But I'm not actually convinced that that is a reason to add this Superstruct, because I think the query parameters case makes more sense as a request-layer middleware. (For example, I've added a Koa middleware to my stack to handle this for me instead.)

If others can provide other good use case examples that would be helpful, otherwise I'm tempted to avoid doing this for complexity's sake.

huanz:
@ianstormtaylor I don't think it's appropriate to do it in the middleware. Because we don't know what parameters are compatible, we can't convert all the string types into numbers in the middleware? Or do you have any good advice?

huanz:
@ianstormtaylor Now I use it to check my request parameters, and I have found another problem.

```javascript
let params = struct(schema, defaults)(this.request.body);
```

If the parameter is not defined in schema , it will be checked into undefined.

But what I want is that the parameters not defined in schema do not make any verification and return directly to me. 

I hope you can think about it.

ianstormtaylor:
@huanz for that I think you might want `struct.partial`. Although that will ignore the non-defined values in the returned object.

lxcid:
> If others can provide other good use case examples that would be helpful, otherwise I'm tempted to avoid doing this for complexity's sake.

For my use case, I'm returning value from Redis, which almost all values are represented as string.
[...]

PR #63, opened 2017-12-20
add numeric type &  strict enum

huanz:
[https://github.com/ianstormtaylor/superstruct/issues/8](https://github.com/ianstormtaylor/superstruct/issues/8)

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -158,7 +158,7 @@ function dict(schema, defaults, options) {
  * @param {Object} options
  */
 
-function en(schema, defaults, options) {
+function en(schema, defaults, options = {strict: true}) {
   if (kindOf(schema) !== 'array') {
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @huanz, thanks for the PR. The way this is currently written this doesn't look like something I'm going to merge though, here's why...

- For the `numeric` type, you can easily add that in your own `superstruct` definition. I think it is too opinionated for core to have.

- For the strict checking, that's not how we'd end up implementing it (since it is opinionated) if we decided to handle coercion—the way this PR implements it only accounts for numbers/strings in enums, which is only one of the many possible use cases. If we decide to do it, it would need to be implemented in a way that is customizable and solves them all nicely.

Hope that makes sense. I'm going to close this, but feel free to follow up.

PR #65, opened 2017-12-20
Allow custom error messages for user defined types

### LINKED ISSUE #60: Allow custom error messages for user defined types
It seems like now is impossible to declare custom error messages for user types

What if in this example I want to provide 2 different error messages, one for first check  `The provided {value} is not email` and other for length check `Email can't be longer than 256 smbs`

```javascript
const struct = superstruct({
  types: {
    email: value => isEmail(value) && value.length < 256,
  }
})
```

May you allow to provide a message in validate function via `throw` or as a result? Or I just not found a solution.

istarkov:
Closes #60

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -326,8 +326,8 @@ const struct = superstruct({
     email: value => isEmail(value) && value.length < 256,
     uuid: value => isUuid.v4(value),
     age: value => {
-      if (value < 16) return "age must be 16 and above"
-      if (value > 125) return "age exceeds theoretical max limit"
+      if (value < 16) return "age_too_small"
+      if (value > 125) return "age_too_big"
       return true;
     }
   }
@@ -342,7 +342,7 @@ const User = struct({
 })
 ```
 
-These custom types are simple functions that return `true/false` or an error message denoting whether the value passed in is valid or not.
+These custom types are simple functions that return `true/false` or a string denoting the reason the value passed in is invalid, in case you want to build more helpful error messages.
 
 
 ## Errors
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -47,8 +47,8 @@ const Struct = struct({
   id: 'number',
   name: 'string',
   is_admin: 'boolean?',
-}, { 
-  is_admin: false 
+}, {
+  is_admin: false
 })
 
 Struct({
@@ -57,7 +57,7 @@ Struct({
 })
 ```
 
-The `struct` function ships with Superstruct by default, and recognizes all of the native JavaScript types in its definitions. It's the easiest way to quickly define structs. 
+The `struct` function ships with Superstruct by default, and recognizes all of the native JavaScript types in its definitions. It's the easiest way to quickly define structs.
 
 If you need to define custom data types, use the [`superstruct`](#superstruct) export instead...
 
@@ -272,7 +272,7 @@ Lazy structs accepts a function that will return a struct. They are useful to cr
 
 Superstruct can be used to validate the structure of data, for things like tuples, dictionaries, lists, etc. But at the lowest level, the data being validate uses type validation functions that you can define yourself.
 
-### Built-in Types 
+### Built-in Types
 
 Out of the box, Superstruct recognizes all of the native JavaScript types:
 
@@ -325,6 +325,11 @@ const struct = superstruct({
   types: {
     email: value => isEmail(value) && value.length < 256,
     uuid: value => isUuid.v4(value),
+    age: value => {
+      if (value < 16) return "age must be 16 and above"
+      if (value > 125) return "age exceeds theoretical max limit"
+      return true;
+    }
   }
 })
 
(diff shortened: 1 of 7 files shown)


docs/reference.md
```diff
@@ -325,23 +325,29 @@ const struct = superstruct({
...
   id: 'uuid',
   name: 'string',
   email: 'email',
+  age: 30,
   is_admin: 'boolean?',
 })
 ```
 
-These custom types are simple functions that return `true/false` denoting whether the value passed in is valid or not.
+These custom types are simple functions that return `true/false` or an error message denoting whether the value passed in is valid or not.
```
ianstormtaylor: Can you change this to refer to it as an "error reason" instead of message? I don't want people to think that putting user-facing messages in these validators is a good idea. Specifically:

> or a string denoting the reason the value passed in is invalid, in case you want to build more helpful error messages.

examples/custom-types.js
```diff
@@ -8,7 +8,11 @@ import isUrl from 'is-url'
 const struct = superstruct({
   types: {
     uuid: v => isUuid.v4(v),
-    email: v => isEmail(v) && v.length < 256,
+    email: (v) => {
+      if (!isEmail(v)) return `${v} is not email`
+      if (v.length >= 256) return 'email length must be less than 256 characters'
```
ianstormtaylor: Can you change these return values to `'not_email'` and `'too_long'`? Again to reinforce the idea that these are machine-readable, and not human-exposed.

src/error.js
```diff
@@ -8,13 +8,16 @@
 class StructError extends TypeError {
 
   constructor(attrs) {
-    const { data, value, type, path, errors = [] } = attrs
+    const { data, value, type, path, code, errors = [] } = attrs
```
ianstormtaylor: I just realized, to be more clear, I think we should call this `reason` instead of `code`. (My fault for saying `code` in the first place.)

src/error.js
```diff
@@ -8,13 +8,16 @@
...
     const message = `Expected a value of type \`${type}\`${path.length ? ` for \`${path.join('.')}\`` : ''} but received \`${JSON.stringify(value)}\`.`
     super(message)
     this.data = data
     this.path = path
     this.value = value
     this.type = type
+    if (code) {
```
ianstormtaylor: Can you change the above to `code = null` (actually `reason = null`) such that we don't have to check here whether it is defined or not, and we can just always set it.

src/kinds.js
```diff
@@ -219,9 +219,14 @@ function func(schema, defaults, options) {
...
-    return schema(value)
+    // can return validation message or isValid
+    const codeOrIsValid = schema(value)
+    const isCode = kindOf(codeOrIsValid) === 'string'
+    const isValid = !isCode && codeOrIsValid
+
+    return isValid
       ? [undefined, value]
-      : [{ type, value, data: value, path: [] }]
+      : [{ type, value, data: value, path: [], ...(isCode ? { code: codeOrIsValid } : undefined) }]
```
ianstormtaylor: Same here, this can be simplified by always passing a `reason` of either the string or `null` if not, so we don't have to do the object mixing in.

test/fixtures/custom/code-invalid.js
```diff
@@ -0,0 +1,22 @@
+
+import isEmail from 'is-email'
+import { superstruct } from '../../..'
+
+const struct = superstruct({
+  types: {
+    email: (v) => {
+      if (!isEmail(v)) return `${v} is not email`
```
ianstormtaylor: Can you change this to be `'not_email'` to match the above?

src/kinds.js
```diff
@@ -219,9 +219,15 @@ function func(schema, defaults, options) {
...
+    // can return string reason or isValid
+    const reasonOrIsValid = schema(value, data)
+    const isReason = kindOf(reasonOrIsValid) === 'string'
+    const isValid = !isReason && reasonOrIsValid
+    const reason = isReason ? reasonOrIsValid : null
+
+    return isValid
       ? [undefined, value]
-      : [{ type, value, data: value, path: [] }]
+      : [{ type, value, data: value, path: [], reason }]
```
istarkov: The only problem with null that now for example

```javascript
const User = struct({
  ttt: struct.enum(['one', 'two'])
})
const data = {}
User(data)
```

will not have { reason:null } in error inside array, because `enum` error is generated in other place.

Do we need to place reason: null there or not, what are you think?

istarkov: PS: I usually prefer `undefined` vs null, because I can easily provide a defaults via object destructuring `({ reason = 'What I want' }) => `

ianstormtaylor: Not exactly sure what you mean, but I'm sold on using `undefined` instead of `null`! The destructuring argument is a good one. We can always change it if really needed.

istarkov: done

ianstormtaylor:
Thanks @istarkov!

Issue #68, opened 2017-12-21
Error nesting structs


doodlewind:
When trying to nest `TreeNode` structs like below I came up with error:

``` js
const Node = struct({
  id: 'number',
  children: [Node]
})
```

[JSFiddle](https://jsfiddle.net/z757neg4/2/)

Maybe I'm missing something, so how can we support such kind of validation? I can such feature pretty handy in slate's schema, is slate going to has this library as dependency?

doodlewind:
Neither does this snippet work:

``` js
let Node
Node = struct({
  id: 'number',
  children: [Node]
})
```

I got following error:

```
superstruct.js:275 Uncaught ReferenceError: process is not defined
    at any (VM82 superstruct.js:275)
    at list (VM82 superstruct.js:550)
    at any (VM82 superstruct.js:234)
    at object (VM82 superstruct.js:633)
    at Object.any (VM82 superstruct.js:244)
    at struct (VM82 superstruct.js:1063)
    at window.onload ((index):66)
```

Related code:

```
  if (process.env.NODE_ENV !== 'production') {
    throw new Error(`A schema definition must be an object, array, string or function, but you passed: ${schema}`);
  } else {
    throw new Error(`Invalid schema: ${schema}`);
  }
```

kenleytomlin:
Hi @doodlewind I think the lazy struct functionality is what you are looking for.

https://github.com/ianstormtaylor/superstruct/blob/master/docs/reference.md#lazy

This looks suitable for your use case.

doodlewind:
@kenleytomlin thanks! That's the missing one I'm looking for.

PR #71, opened 2017-12-21
Strip out NODE_ENV in UMD build

doodlewind:
In #68 following error thrown in JSFiddle: 

```
superstruct.js:275 Uncaught ReferenceError: process is not defined
    at any (VM82 superstruct.js:275)
    at list (VM82 superstruct.js:550)
    at any (VM82 superstruct.js:234)
    at object (VM82 superstruct.js:633)
    at Object.any (VM82 superstruct.js:244)
    at struct (VM82 superstruct.js:1063)
    at window.onload ((index):66)
```

This is due to `process.env.NODE_ENV` is preserved in UMD build, keeping this variable in browser leads to an error.

To my understanding, following strategy is ideal for `NODE_ENV`:

* [x] When used in Node.js server side, `process.env.NODE_ENV` should be kept.
* [x] When used in module bundler, i.e., webpack, `process.env.NODE_ENV` should be kept, leaving user to strip them with things like `DefinePlugin`.
* [ ] When used directly in browser, `process.env.NODE_ENV` should be stripped when we build the lib. It's reasonable to show dev log when using `umd/superstruct.js`, and production warning with `umd/superstruct.min.js`.

So this PR implements the last missing point. With `rollup-plugin-replace` we can easily achieve this without modifying code in `src`.

### DIFF
--- a/config/rollup-umd.js
+++ b/config/rollup-umd.js
@@ -1,5 +1,10 @@
 
 import config from './rollup'
+import replace from 'rollup-plugin-replace'
+
+config.plugins.push(
+  replace({ 'process.env.NODE_ENV': JSON.stringify('dev') })
+)
 
 config.output = {
   file: './umd/superstruct.js',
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Perfect, thanks @doodlewind!

PR #73, opened 2017-12-22
Fix doc catalog href

doodlewind:
This PR ensures all API under `#structs` links to existing item in alphabetical order.

P.S. There are still some missing `Structs` validator missing in reference manual. Are these kinds designed to be private? I did a quick search and have the following list for missing items:

- [ ] any
- [ ] enums
- [ ] instance
- [ ] interface
- [ ] literal
- [ ] partial
- [ ] tuple

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -11,11 +11,11 @@
   + [`enum`](#enum)
   + [`function`](#function)
   + [`intersection`](#intersection)
+  + [`lazy`](#lazy)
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thank you @doodlewind! 

Those other ones aren't private, I just forgot to add them to the docs originally and haven't had time to revisit it. But if anyone wants to that would be awesome!

PR #74, opened 2017-12-22
Fix version-badge link

kvartborg:


ianstormtaylor:
Ah oops, thanks @kvartborg!

PR #87, opened 2018-02-03
Add `micro-superstruct` to resources.md

brandon93s:
Thank you for the awesome validation library! I've added `micro-superstruct` to resources which allows you to use Superstruct for API validation with Micro.


ianstormtaylor:
Cool, thank you @brandon93s!

Issue #89, opened 2018-02-07
Expected a value of type `undefined` for `author.test` but received `undefined`


saighost:
```
'use strict';
const { struct } = require('../lib/superstruct/lib/index');
const Article = struct({
  id: 'number',
  author: {
    id: 'number',
  },
});
const data = {
  id: 34,
  author: {
    id: 1,
    test: undefined,
  },
};
const article = Article(data);
console.log('article', article);
```
I use the struct to validate data from excel, some unsed keys are set to undefined, and I get an error like :
Expected a value of type `undefined` for `author.test` but received `undefined`

I try to partial, but it is not recursive. so I create this issues and a pull request.

szhangpitt:
I think it could also be a matter of the error message... e.g. the message could be “didn’t expect a property ‘test’ but got ‘test: undefined’” ... instead of the current one. 

Having a property with value of undefined vs. not having this property at all is two concept; how this module treat them is a design decision we can then discuss.

saighost:
Undefined is not a null.
This is a data validator, so I care about the value when the key in the schema. And it validates the type of value.
This issue wants to ignore keys these not in the schema and with an undefined value. The only difference is these keys is enumerable.

szhangpitt:
If I understand it correctly, you would like your code in [first comment](https://github.com/ianstormtaylor/superstruct/issues/89#issue-295298067) to pass the check and not throw error?

Did you try this? (Seems to work on 0.5.0)

```js
const {struct} = require('superstruct');

const Article = struct({
  id: 'number',
  author: struct.partial({      // use struct.partial at this level
    id: 'number',
  }),
});
const data = {
  id: 34,
  author: {
    id: 1,
    test: undefined,
  },
};

const article = Article(data);
console.log('article', article);
```

szhangpitt:
You suggested to make `struct.partial()` recursive; I think that's a totally valid request. However I think it will change the essential behavior of `struct.partial()` (I know it's not 1.0.0 yet, but still..) 

My understanding for this library is that, object based schema is defined as explicit as possible, and any unexpected key (enumerable as you mentioned) will raise an error; that's more of a design decision. Another feature of this library is being highly compose-able, so that on each level, you can pass in a `struct` object to do more refined validation. 

To make `struct.partial()` recursive, I'm afraid it's somewhat going into opposite direction of compose-ability. 

This is only my opinion; apologize if I understood your intention incorrectly; and I'd love @ianstormtaylor  's insight.

[...]

PR #90, opened 2018-02-07
Two way for #89

saighost:
#89 
made partial and interface to apply recursively
ignore the object key with an undefined value and without a default value.


ianstormtaylor:
Hey @saighost can you turn off the formatting stuff? I think I like the direction of this but it's hard to tell with the formatting noise in the PR. Thanks!

saighost:
sorry, I will fix it,

src/kinds.js
```diff
@@ -285,7 +285,12 @@ function inter(schema, defaults, options) {
   for (const key in schema) {
     ks.push(key)
     const s = schema[key]
-    const kind = any(s, undefined, options)
+    let kind
+    if (kindOf(s) === 'object') {
+      kind = inter(s, undefined, options)
+    } else {
+      kind = any(s, undefined, options)
+    }
```
ianstormtaylor: Hey @saighost would you mind changing these lines to be ternaries, like:

```
const kind = kindOf(s) === 'object' ? ...
```

src/kinds.js
```diff
@@ -552,7 +559,12 @@ function partial(schema, defaults, options) {
   for (const key in schema) {
     ks.push(key)
     const s = schema[key]
-    const kind = any(s, undefined, options)
+    let kind
+    if (kindOf(s) === 'object') {
+      kind = partial(s, undefined, options)
+    } else {
+      kind = any(s, undefined, options)
+    }
```
ianstormtaylor: Hey @saighost would you mind changing these lines to be ternaries, like:

```
const kind = kindOf(s) === 'object' ? ...
```

src/kinds.js
```diff
@@ -483,8 +488,10 @@ function object(schema, defaults, options) {
         const d = defaults && defaults[key]
         v = resolveDefaults(d, value)
       }
-
       if (!kind) {
+        if (v === undefined) {
+          return
+        }
```
ianstormtaylor: What's this extra check needed for?

saighost: oh, for a key with an undefined value.
My data is wrapped by Proxy to filter and format value, and when an unknown key is appeared, I can only return an undefined value.

saighost: I try partial, but it can't ignore children's properties.

ianstormtaylor: Could you open a separate issue/PR for this so we can discuss it? It seems unrelated to the partial/object issue.

src/kinds.js
saighost: Finish.

src/kinds.js
saighost: #89 has a small test case.
I can try to add a test case for it.

ianstormtaylor:
Hey @saighost, thank you for this PR. I believe that as of #139 with the changes to `struct.interface` and `struct.partial` this makes less sense now, so I'm going to close it. But if you disagree, feel free to reply!

PR #95, opened 2018-02-22
Add static 'format' method on StructError

### LINKED ISSUE #94: Expose `StructError` message formatter directly
It'd be nice to have access to the formatter directly. It can be used within (and exported from) the same file to create the `message`.

Currently, if you have multiple errors and want to format them, you'd have to go thru the "work" on instantiating a new class per error, just to take the message string.

```js
let messages = err.errors.map(attr => (new StructError(attr)).message);

// vs

function toMessage(o) {
  return `Expected a value of type \`${o.type}\`${o.path.length ? ` for \`${o.path.join('.')}\`` : ''} but received \`${JSON.stringify(o.value)}\`.`;
}

let messages = err.errors.map(toMessage);
```

lukeed:
This closes #94 as suggested, but I think I'd prefer my initial approach since this is a bit hacky & still requires importing the entire class object to access the formatter.

It's your call, but ideally, I'd do this:

```js
export function formatError(attrs) {
  // ...same
}

export class StructError extends TypeError {
  constructor(attrs) {
    const msg = formatError(attrs)
    super(msg)
    const errors = attrs.errors || []
    // ...same
  }
}

// ---

import { formatError, StructError } from './error'

export {
  struct,
  superstruct,
  isStruct,
  formatError,
  StructError
}
```

```js
import { formatError } from 'superstruct';
```

### DIFF
--- a/src/error.js
+++ b/src/error.js
@@ -6,6 +6,7 @@
  */
 
 class StructError extends TypeError {
+
   constructor(attrs) {
     super()
     const errors = attrs.errors || []
@@ -25,9 +26,10 @@ class StructError extends TypeError {
   }
 
   static format(attrs) {
-    let path = attrs.path.join('.')
+    const path = attrs.path.join('.')
     return `Expected a value of type \`${attrs.type}\`${path.length ? ` for \`${path}\`` : ''} but received \`${JSON.stringify(attrs.value)}\`.`
   }
+
 }
 
 /**
--- a/src/error.js
+++ b/src/error.js
@@ -6,16 +6,15 @@
  */
 
 class StructError extends TypeError {
-
   constructor(attrs) {
-    const { data, value, type, path, reason, errors = [] } = attrs
-    const message = `Expected a value of type \`${type}\`${path.length ? ` for \`${path.join('.')}\`` : ''} but received \`${JSON.stringify(value)}\`.`
-    super(message)
-    this.data = data
-    this.path = path
-    this.value = value
-    this.type = type
-    this.reason = reason
+    super()
+    const errors = attrs.errors || []
+    this.message = this.constructor.format(attrs)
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
What's wrong with importing "the entire" `StructError` object?

lukeed:
Nothing :) Just preference

src/error.js
```diff
@@ -8,14 +8,14 @@
 class StructError extends TypeError {
 
   constructor(attrs) {
-    const { data, value, type, path, reason, errors = [] } = attrs
-    const message = `Expected a value of type \`${type}\`${path.length ? ` for \`${path.join('.')}\`` : ''} but received \`${JSON.stringify(value)}\`.`
-    super(message)
```
ianstormtaylor: Why remove the `message` in the call to `super`? I'm not sure if that's okay or not for inheriting errors. Do you have proof that it's fine?

src/error.js
```diff
@@ -8,14 +8,14 @@
 class StructError extends TypeError {
 
   constructor(attrs) {
-    const { data, value, type, path, reason, errors = [] } = attrs
```
ianstormtaylor: Why change this logic?

lukeed: It's just adding bytes since all of these keys are only used once. The only thing _needed_ was fallback for `errors`.

src/error.js
lukeed: You don't have access to `this` until `super()` is called. Passing any first argument to the base `Error` constructor sets it as the `err.message` property, so we just set it immediately after since it's always writable.

ianstormtaylor:
Fair enough, thanks @lukeed.

lukeed:
🙌 Thanks @ianstormtaylor

Issue #99, opened 2018-03-09
contains solution with unit test: custom validator's returned 'error'(reason) string not consistently returned


llpage30:
I have a data type 'one' that I want to return the value of the string if that value is not of length 1.
The StructError.reason, in these cases, should then be the string I returned.

The issue I am hitting and is eventually shown by the code below, is if I wrap my object definition containing this custom validator in an 'optional' call, the returned string 'gets lost' resulting in StructError.reason === undefined.  
This is a painful bug, and I cannot determine any work around for this.
Please fix this.

```javascript
const { superstruct } = require('superstruct');
const assert = require('assert');
// our 'one' type that should return the value in the structError.reason if value length != 1.
const TYPES = superstruct({
	types: {
		one: (value) => {
			if(value.length === 1) return true;
			return value;
		},
	},
});
const INNER = TYPES.partial({
	id: 'one',
	name: 'string',
});
// The use case I have is with a nested definition. These all work....
const OUTER_WORKS1 = TYPES.partial({
	no: 'one',
	user: INNER,
});
const OUTER_WORKS2 = TYPES.partial({
	no: 'one',
	user: [INNER],
});
// I wanted to have that nested definition be optional, but can't as the reason string does not come through.....
const OUTER_FAILS1 = TYPES.partial({
	no: 'one',
	user: TYPES.optional(INNER),
});
const OUTER_FAILS2 = TYPES.partial({
	no: 'one',
	user: TYPES.optional([INNER]),
});

// some data objects that will use in the asserts...
const innerDataGood = { id: '1', name: 'hello' };
const innerDataBad = { id: '11', name: 'hello' };
const outerDataGood = { no: 'y', user: innerDataGood };
const outerDataBad1 = { no: 'yes', user: innerDataGood };
const outerDataBad2 = { no: 'y', user: innerDataBad };
const outerData2Good = { no: 'y', user: [innerDataGood] };
const outerData2Bad1 = { no: 'yes', user: [innerDataGood] };
const outerData2Bad2 = { no: 'y', user: [innerDataBad] };

/**
  This all works in this function.
   This is here to indicate the structures above are valid and do return the reason string I wanted
   note: non of the FAILS cases are in this function.
**/
const baselineVerifyData = () => {
	let goodResult = INNER.validate(innerDataGood);
	assert.deepEqual(goodResult[1], innerDataGood);

	goodResult = OUTER_WORKS1.validate(outerDataGood);
[...]

PR #101, opened 2018-03-28
Fix error reason on optional custom types

Barterr:
Fixing #99 
Reason for optional custom types was always undefined
Added a test for that case


ianstormtaylor:
Thank you @Barterr, that is awesome! Great PR.

tomspeak:
Thanks a lot @Barterr & @ianstormtaylor.

Issue #102, opened 2018-04-26
Customize Error Paths


davidchase:
Hi 👋 really digging this library for validating objects

My only question is about customizing paths when creating types via superstruct

```js
const struct = superstruct({
  types: {
    location: value => {
      const empty = Object.keys(value).reduce((acc, key) => {
        return value[key].state === null && value[key].county === null
          ? acc.concat(key)
          : acc;
      }, []);
      const str = empty.join(", ");
      return str === ""
        ? true
        : `Either state or county for ${str} is required`;
    },
    duplicate: val => {
      const mmap = Object.keys(val).reduce((map, key) => {
        CHECK_LIST.forEach(item => {
          if (val[key][item] !== "" && val[key][item] != null) {
            map.set(`${item}.${val[key][item]}`, `${key}.${item}`);
          }
        });
        return map;
      }, new MultiMap());
      
      const gatherDuplicates = mmap
        .values()
        .reduce((acc, x) => (x.length >= 2 ? acc.concat(x) : acc), []);
        
      const error = `The following keys are duplicates: ${gatherDuplicates.join(", ")}`

      return gatherDuplicates.length > 0 ? error : true;
    }
  }
});

const Location = struct('location&duplicate')

const data = {
  123: {
    state: null,
    county: null
  },
  456: {
    state: null,
    county: null
  }
}

Location.validate(data)
```

Currently the above code checks if the dynamic keys have either state or county filled in and then if either state or county is a duplicate

Returning custom messages works great but for whatever reason the `paths` array is always empty which brings me to my question of customizing paths
because i have all of the information i need to create a path but no obvious way of doing, unless I am missing something ?

davidchase:
@ianstormtaylor thanks for the quick response 

> Returning an object is interesting. I’d be down for that if you wanted to make a PR.

Yes i can definitely do a PR

> I’d generally try to avoid these kinds of setups, but if two values are co-dependent then there’s no other way really.

I'm open to other ideas, originally i was thinking about returning an array like `['message', ['paths']]` but then developers have to remember
which goes first message or array of paths. Or if you think their might be a better solution altogether im up for discussion.

ianstormtaylor:
@davidchase what's the data shape you're trying to assert? It's a bit hard to tell from the code sample above what "correct" data can look like.

ianstormtaylor:
Interesting!

So what if you structured it as such:

```js
const State = struct.object({ 
  state: 'string', 
  county: 'string|null' 
[...]

PR #105, opened 2018-04-28
Return an { message, path } object from custom types

davidchase:
ref #102


panva:
Maybe a case for `string` can still be kept so that this is not a breaking change? (I know while below ^1.0.0 anything minor can still be breaking, but still, not too much effort to support all three cases (string, boolean and object).

davidchase:
@panva thats a good point, we should keep string as a backwards compatibility option... I didn't even realize superstruct was not v1

package.json
```diff
@@ -53,7 +53,7 @@
     "build:max": "rollup --config ./config/rollup-umd.js",
     "build:min": "rollup --config ./config/rollup-umd-min.js",
     "clean": "rm -rf ./lib ./umd ./node_modules",
-    "lint": "eslint '{src,test}/*' && prettier --list-different '**/*.{js,json,md}'",
+    "lint": "eslint \"{src,test}/*\" && prettier --list-different \"**/*.{js,json}\"",
```
davidchase: i have windows work machine so i need to wrap the eslint commands in double quotes to get it to work. if this is a problem i will remove

.eslintrc
```diff
@@ -37,7 +37,7 @@
     "import/no-named-as-default": "error",
     "import/no-named-as-default-member": "error",
     "import/no-unresolved": "error",
-    "linebreak-style": "error",
+    "linebreak-style": ["error", "windows"],
```
davidchase: same as below.. windows made me do it 😞 lol

src/kinds.js
```diff
@@ -232,25 +232,36 @@ function func(schema, defaults, options) {
   const type = '<function>'
   const validate = (value = resolveDefaults(defaults), data) => {
     const result = schema(value, data)
```
davidchase: the logic was getting hairy keeping track of all of the supported types so i refactored a little bit let me know if it works

davidchase:
ahh travis took a dive because of the eslint `line-breaks` differ on windows and linux/mac osx

ianstormtaylor:
Thanks @davidchase!

PR #106, opened 2018-04-30
dict, interface defaults subset

### LINKED ISSUE #104: dict and intersection defaults
Hi @ianstormtaylor,

i have the following use cases for which i've created [tests](https://gist.github.com/panva/9f6239a135ed9436683f749ab0586eb3), that currently fail

1) partial-defaults.js

Shouldn't keys from defaults be returned if not found in data?

2) defaults-partial-dict-one.js and defaults-partial-dict-two.js

Two ways I thought would work to achieve 1) but they're both failing in different ways.

```
  1) superstruct
       dict-scalars
         partial-defaults:

      AssertionError [ERR_ASSERTION]: { b: 3, c: 4 } deepEqual { a: 1, b: 3, c: 4 }
      + expected - actual

       {
      +  "a": 1
         "b": 3
         "c": 4
       }
      
      at Context.it (test/index.js:33:20)

  2) superstruct
       intersection
         defaults-partial-dict-1:

      AssertionError [ERR_ASSERTION]: { a: 3, b: 2 } deepEqual { a: 3, b: 2, c: 5 }
      + expected - actual

       {
         "a": 3
         "b": 2
      +  "c": 5
       }
      
      at Context.it (test/index.js:33:20)

  3) superstruct
       intersection
         defaults-partial-dict-2:
     TypeError: Expected a value of type `dict<string,number> & {a,b,...}` for `b` but received `undefined`.
      at Function.Struct.assert.value [as assert] (lib/index.js:1119:15)
      at Struct (lib/index.js:1103:21)
      at Context.it (test/index.js:32:28)
```

panva:
dict and interface defaults subset fix
interfaces composed of structs with their own defaults now resolve those defaults
Also adds .only and .skip support for individual tests (by exports)

I think it closes #104

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -313,17 +313,28 @@ function inter(schema, defaults, options) {
     const resolved = resolveDefaults(defaults)
     value = resolved ? { ...resolved, ...value } : value
     const errors = []
+    const ret = value
 
     for (const key in properties) {
-      const v = value[key]
+      let v = value[key]
       const kind = properties[key]
-      const [e] = kind.validate(v)
+
+      if (v === undefined) {
+        const d = defaults && defaults[key]
+        v = resolveDefaults(d, value)
+      }
+
+      const [e, r] = kind.validate(v, value)
 
       if (e) {
         e.path = [key].concat(e.path)
         e.data = value
         errors.push(e)
         continue
+      }
+
+      if (key in value || r !== undefined) {
+        ret[key] = r
       }
     }
 
@@ -333,7 +344,7 @@ function inter(schema, defaults, options) {
       return [first]
     }
 
-    return [undefined, value]
+    return [undefined, ret]
   }
 
   return new Kind(name, type, validate)
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -109,7 +109,9 @@ function dict(schema, defaults, options) {
   const values = any(schema[1], undefined, options)
   const name = 'dict'
   const type = `dict<${keys.type},${values.type}>`
-  const validate = (value = resolveDefaults(defaults)) => {
+  const validate = value => {
+    const resolved = resolveDefaults(defaults)
+    value = resolved ? { ...resolved, ...value } : value
     const [error] = obj.validate(value)
 
     if (error) {
@@ -307,7 +309,9 @@ function inter(schema, defaults, options) {
 
   const name = 'interface'
   const type = `{${ks.join()}}`
-  const validate = (value = resolveDefaults(defaults)) => {
+  const validate = value => {
+    const resolved = resolveDefaults(defaults)
+    value = resolved ? { ...resolved, ...value } : value
     const errors = []
 
     for (const key in properties) {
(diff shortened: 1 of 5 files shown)


src/kinds.js
```diff
@@ -109,7 +109,9 @@ function dict(schema, defaults, options) {
   const values = any(schema[1], undefined, options)
   const name = 'dict'
   const type = `dict<${keys.type},${values.type}>`
-  const validate = (value = resolveDefaults(defaults)) => {
+  const validate = value => {
+    const resolved = resolveDefaults(defaults)
+    value = resolved ? { ...resolved, ...value } : value
```
ianstormtaylor: Hey @panva question about these lines (in both places). `value` has the same potential to be `undefined/null` as `resolved` does, but there is only a guard for the case of `resolved` being `undefined/null`?

Shouldn't it be that we either don't need a guard at all (which is great), or that we need a guard for each?

test/index.js
```diff
@@ -23,10 +23,10 @@ describe('superstruct', () => {
         .map(t => basename(t, extname(t)))
 
       for (const test of tests) {
-        it(test, () => {
-          const module = require(resolve(testsDir, test))
-          const { Struct, data } = module
-
+        const module = require(resolve(testsDir, test))
+        const { Struct, data, only, skip } = module
+        const run = only ? it.only : skip ? it.skip : it
```
ianstormtaylor: Great call here!

src/kinds.js
panva: But only when defaults return something do you actually want to merge the two, else you return an empty unexpected object as value.

ianstormtaylor: Ah, great call!

src/kinds.js
```diff
@@ -307,20 +309,33 @@ function inter(schema, defaults, options) {
...
     for (const key in properties) {
-      const v = value[key]
+      let v = value[key]
       const kind = properties[key]
-      const [e] = kind.validate(v)
+
+      if (v === undefined) {
+        const d = defaults && defaults[key]
+        v = resolveDefaults(d, value)
+      }
```
ianstormtaylor: Can you explain what this part is doing? Why would defaults need to be resolved twice here? And isn't accessing `defaults[key]` potentially incorrect since defaults can be a function?

src/kinds.js
```diff
@@ -307,20 +309,33 @@ function inter(schema, defaults, options) {
...
 
       if (e) {
         e.path = [key].concat(e.path)
         e.data = value
         errors.push(e)
         continue
       }
+
+      if (key in value || r !== undefined) {
+        ret[key] = r
```
ianstormtaylor: Just to be sure, is this logic here so that the return type is still the same, since interfaces can be non-plain objects?

src/kinds.js
panva: I actually took the code from `partial` for this part. It was the only way i could make referenced structs with their defaults re-usable inside interfaces. See the test added below.

src/kinds.js
panva: Same as above, just aligned the code between `partial` and `interface`

src/kinds.js
ianstormtaylor: Ah cool, thanks @panva!

ianstormtaylor:
Thanks for taking the time to write this up @panva!

Issue #111, opened 2018-08-02
List struct only returns first error when validated


toxsick:
When using list structs I world expect that errors.error includes all validation errors for all list items. Instead it seems that only the first encountered error is returned.

```javascript
import { struct } from 'superstruct';

const ListTest = struct({
  list: struct.list(['string'])
});

const data = {
  list: [1, 1]
};

const [error] = ListTest.validate(data);
console.log(error.errors.map(e => e.path));

// shound be
// [ [ 'list', 0 ], [ 'list', 1 ] ]

// actual
//[ [ 'list', 0 ] ]
```

I don't think there is a testcase for that also.

regards
Hannes

toxsick:
By looking a bit more at the tests I think `error.errors` is not tested at all currently. Extending the general test logic for this would be overkill. I try to find the root issue, but directions on how to test this would be helpful.

thx and regards
Hannes

toxsick:
Thanks Ian!

PR #112, opened 2018-08-02
pass all errors of composed structs for #111

toxsick:
#111

Currently only the first error of a sub struct (e.g. list) is included in the parent object struct.

TDB if there are more implications for other struct kinds. Further the testsuite has to be extended for this.

cheers

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -127,9 +127,12 @@ function dict(schema, defaults, options) {
       const [e, r] = keys.validate(k)
 
       if (e) {
-        e.path = [k].concat(e.path)
-        e.data = value
-        errors.push(e)
+        const allE = e.errors || [e]
+        allE.forEach(singleE => {
+          singleE.path = [k].concat(singleE.path)
+          singleE.data = value
+          errors.push(singleE)
+        })
         continue
       }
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @toxsick, thanks for this! It looks like you've edited the `object` struct instead of the `list` struct? Maybe you meant to do both?

toxsick:
Hi Ian,

first of all thanks for this library, it is really fun to work with! I tried to apply this logic to all existing struct kinds now (my first shot only fixed my particular use case). A few things that I need feedback on:

1. Do you like the naming of the variables I introduced: `allE` `singleE` If not give me a suggestion and I will change them.
2. I also want to include tests for this, but this is a lot of work since the current test logic checks if the fixtures deepEqual to result. Any suggestions on how we could include this in the testuite?
3. I'm not sure how to deal with intersections (https://github.com/ianstormtaylor/superstruct/blob/master/src/kinds.js#L846-L848) Do I have to add deep error support here too? I think maybe not..

Give me feedback on this and I will see what I can do since really would like to get this merged for a project of mine.

regards
Hannes

toxsick:
Hey @ianstormtaylor,

I know u r a busy guy. But do you perhaps have an update for me in this?

regards
Hannes

ianstormtaylor:
Hey @toxsick, I'm sorry for the delay here!

This looks great, I might refactor the logic into a helper function to save on repetition, but I can do that in a future commit if needed. I think you're right that with intersections we don't need deep error support across the intersections, and just the first one that failed would be the one to go with.

Thank you so much for writing this!

PR #118, opened 2018-11-07
intersection struct should contain all errors of intersected child structs

toxsick:
In the spirit of my previous PR #112 I came across an issue with the intersection struct. Currently it will only return the errors of the first child struct that throws an error. I changed the intersection struct to return all errors now.

This is still not covered by the testsuite though, since the returned errors are not part of it.

Tell me what you think.

cheers

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -127,9 +127,12 @@ function dict(schema, defaults, options) {
       const [e, r] = keys.validate(k)
 
       if (e) {
-        e.path = [k].concat(e.path)
-        e.data = value
-        errors.push(e)
+        const allE = e.errors || [e]
+        allE.forEach(singleE => {
+          singleE.path = [k].concat(singleE.path)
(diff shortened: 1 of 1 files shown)


toxsick:
@ianstormtaylor did u find the time to have a look at this?

toxsick:
New year new try, any updates on this @ianstormtaylor ?

Thanks and happy new year

ianstormtaylor:
Hey @toxsick sorry for the silence here. Can you add a test to ensure this continues to work in the future? Thanks.

toxsick:
Hey @ianstormtaylor, sorry but I missed this one. As I stated initally the problem is that the testsuite does not check the errors that are returned. This has not changed right?

ianstormtaylor:
@toxsick it doesn't currently, but it can be an opt-in if the expected object includes a `failure` property, which should work in this case.

PR #122, opened 2019-01-08
fix nest errors struct

teenth:
Error message structure nested in each other

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -154,7 +154,7 @@ function dict(schema, defaults, options) {
 
     if (errors.length) {
       const first = errors[0]
-      first.errors = errors
+      first.errors = JSON.parse(JSON.stringify(errors))
       return [first]
     }
 
@@ -362,7 +362,7 @@ function inter(schema, defaults, options) {
 
     if (errors.length) {
       const first = errors[0]
-      first.errors = errors
+      first.errors = JSON.parse(JSON.stringify(errors))
       return [first]
     }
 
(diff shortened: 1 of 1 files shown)


pckilgore:
Hey @ianstormtaylor thanks so much for the library.  Superstruct + gql introspection queries might just be the best blend of type driven composition/validation I've ever worked with on the frontend.

I wanted to raise the profile of this PR since it's been sitting a few months.  It really improves the error messages we see, particularly the developer experience through Sentry etc.   

If you're not fond of `stringify(parse(`,  it doesn't seem to be a real world performance hit.  We're using these changes on a rather large input object with multiple nest array field input types from our gql API it validation is lightning fast.  You can also try (fast-json-stringify)[https://github.com/fastify/fast-json-stringify].

Thanks again!

ianstormtaylor:
Hey @teenth thank you for addressing this! And @pckilgore thanks for the bump.

I'm worried about the `stringify(parse(` because it also means that non-JSON references can't be handled here. I'd be open to an alternative solution that doesn't impose this restriction.

I didn't look closely, but what if we turned:

```js
const first = errors[0]
first.errors = errors
return [first]
```

Into:

```js
const first = { ...errors[0], errors }
return [first]
```

(Didn't test this out, could be pseudocode 😄)

pckilgore:
@ianstormtaylor Thanks! I'll investigate either tomorrow or over the weekend and see how it goes.

pckilgore:
@ianstormtaylor Works!  PR implementing @ #126.

PR #124, opened 2019-03-25
Dynamic structs

DecentM:
I've come across a use case where I needed to validate many possible structs in a tree-like data structure.  
I've managed to accomplish that by abusing the custom type system:

<details><summary>Code example <b>(click to expand)</b></summary>

```javascript
import {superstruct, struct} from 'superstruct'

const base = struct({'someValue': 'boolean'})
export const Struct = superstruct({
  'types': {
    'base': base.test,
  }
})('base')
```

</details>

While this works as far as validation of the correctness is concerned, I lose the awesome error messages Superstruct generates. This is because the error will contain the fact that only the topmost element in my data tree is incorrect. I'd expect the path to drill down to the lowest level and then work its way up the tree.  

This PR contains a non-breaking change to `src/kinds.js` that makes it possible to fix this behaviour. I've also included tests for the lazy struct in general (which seem to be missing from `test/fixtures`) and the described issue above.  

Thanks for this project, it's saved me (and I'm sure many others) a bunch of hours of rooting around in data!

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -259,6 +259,60 @@ const BinaryTree = struct({
 
 `lazy` structs accepts a function that will return a struct. They are useful to create recursive structs.
 
+### `dynamic`
+
+```js
+const map = {
+  TEXT: struct({
+    content: 'string'
+  }),
+  IMAGE: struct({
+    url: 'string'
+  })
+}
+
+const Node = struct({
+  type: struct.enum(['TEXT', 'IMAGE']),
+  value: struct.dynamic((value, parent) => {
+    // Make sure your function never comes back with anything but a struct.
+    return map[parent.type] || struct('undefined')
+  })
+})
+
+const Struct = struct({
+  nodes: [Node]
+})
+```
+
+```js
+{
+  nodes: [
+    {
+      type: 'TEXT',
+      value: {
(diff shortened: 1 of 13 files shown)


ianstormtaylor:
Hey @DecentM, thank you for taking the time to work on this!

Your use case seems like a reasonable one to me. 

The only thing is that I think the `lazy` kind should have actually been written to only ever evaluate its schema function once, the first time it is run. (This would currently be a bug then.)

And instead, maybe we can add a `dynamic` kind as well? Which has the behavior you’re after. Are you down to make that change?

DecentM:
I've tried using my branch experimentally and came across the exact bug you're describing. I think it would make sense to create a new kind. I'm totally down, I'll see if I can do that instead.  

In the meantime, I think I'll revert the change to the `lazy` kind. The new functionality will be on a new `dynamic` kind, and this behaviour is unexpected and makes using dynamic lazy structs hard.

DecentM:
Alright, @ianstormtaylor I think it's ready for review.  

Also, I left the tests for `lazy` in for now, would you prefer that in a different PR?

ianstormtaylor:
Thank you @DecentM, looks great! And the extra tests for `lazy` are very welcome 😄

PR #125, opened 2019-04-04
add failing intersection fixture

erquhart:
`intersection` is a critical function for robust schema validation, but it doesn't seem to be working properly. I'm happy to help with whatever issues might be at play here, but wanted to confirm my expectations first.

This PR adds a failing fixture that represents how I understand the `intersection` struct to work in the docs. Below is the test and it's current output:

```js
export const Struct = struct.intersection([
  struct.partial({
    name: 'string',
  }),
  struct.partial({
    age: 'number',
  }),
])

export const data = {
  name: 'john',
  age: 42,
}

export const output = {
  name: 'john',
  age: 42,
}
```

Output:
```
 1) superstruct
       intersection
         valid:
     TypeError: Expected a value of type `{name,...} & {age,...}` for `age` but received `undefined`.
```

It seems to be expecting the entire combined `intersection` struct to validate the value of `age` rather than the entire object, and it's reading the value of `age` as `undefined`.

Am I doing this right?

### DIFF
--- a/test/fixtures/intersection/valid.js
+++ b/test/fixtures/intersection/valid.js
@@ -1,0 +1,20 @@
+import { struct } from '../../..'
+
+export const Struct = struct.intersection([
+  struct.partial({
+    name: 'string',
+  }),
+  struct.partial({
+    age: 'number',
+  }),
+])
+
+export const data = {
+  name: 'john',
+  age: 42,
+}
+
+export const output = {
+  name: 'john',
+  age: 42,
+}


ianstormtaylor:
@erquhart hmm that is strange. I'm not sure exactly what's going on, but it does sound like a bug to me. I'd love a fix if you can manage it!

I'm kind of unsure how `partial` should work in an `intersection/union`, since it masks the return value. `interface` is more clear, but doesn't have the masking. (Maybe this library should never mask?)

erquhart:
Gotcha. I probably won't be able to put in a fix myself, but I can say for whoever does (if at all) that `partial` was the only way I could find to require that an object have a set of keys while still allowing the object to have unknown keys. In this case I would expect intersection to require that all structs match, so an object would need to have string "name" and number "age".

ianstormtaylor:
@erquhart did using `struct.interface` not work?

erquhart:
I attempted to convert all of our config validation over to superstruct and threw quite a lot of complicated scenarios at it, but it's been a while and I don't remember a lot of specifics. I do remember using `interface()` and I don't recall it being the ticket for this particular issue.

This is a fantastic lib, really wish I had time to invest! Between this, Slate, and the other stuff you're undoubtedly working on, I really don't know how you do it! Feel free to close this if you'd rather keep your issues fresh and I can open something new if I get to give superstruct another spin 👍

PR #126, opened 2019-05-10
Provides better error messages for nested errors

pckilgore:
Previously nested errors messages would describe the parent type of a nested error, rather than the type of the field that actually failed to validate.  This isn't exactly inaccurate (a parent is invalid if a child is invalid), but not very helpful for identifying errors in these circumstances.  

Example previously:

> Error message: "Expected a value of type `[{SectionType1,SectionType2,SectionType3} | undefined | null] | undefined | null` for `sections.0.blockInputTypeA.0.notes` but received `undefined`."

The equivalent error now reads:

> Error message: "Expected a value of type `String` for `sections.0.blockInputTypeA.0.notes` but received `undefined`."

Which more accurately reports the actual field that failed to validate.

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -152,13 +152,7 @@ function dict(schema, defaults, options) {
       ret[k] = r2
     }
 
-    if (errors.length) {
-      const first = errors[0]
-      first.errors = errors
-      return [first]
-    }
-
-    return [undefined, ret]
+    return errors.length ? [{ ...errors[0], errors }] : [undefined, ret]
   }
 
   return new Kind(name, type, validate)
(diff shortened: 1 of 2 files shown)


pckilgore:
:flags: I had to fix a test that was depending on the old error type string.  I think the new error is superior for all the same reasons, but now nervous the old spec might be depended upon, somewhere.

pckilgore:
👋  @ianstormtaylor

ianstormtaylor:
Thanks @pckilgore! Sorry for the delay here.

pckilgore:
@ianstormtaylor no problem, thanks again for maintaining the lib!

Issue #83, opened 2018-01-16
Error when bundle with webpack


forsigner:
Unexpected token: name (StructError) [./node_modules/superstruct/lib/index.es.js:7,0][index.ef29450fbb2e4dc5d2e2.js:147554,6]

doodlewind:
Are you using a dev build or a production one? In production, if you don't translate ES6 in `node_modules`, uglify plugin may throw such error since our bundle contains dependency in ES6 while this plugin can only transform ES5. If using ES6 in production is fine for you, using `uglify-es` can be a workaround.

forsigner:
In my case, I use superstruct with `npm i superstruct -S`.

doodlewind:
Possible same issue: https://github.com/webpack/webpack/issues/1542#issuecomment-308747244

@ianstormtaylor how about releasing `index.es.js` in ES5?

ianstormtaylor:
@doodlewind open to the idea!

Is the issue just with the `import` syntax Babel transforms? I think it would make sense for us to not use `import` in this library, and change them to `require` for CJS support instead. What do you think? Or is it something else?

jackmellis:
I'm having this issue too. I think the distributed files should be ES5 compatible (not just commonjs). Right now I have to write an awkward exclude regex in my webpack config to include superstruct in my babel parsing

PR #128, opened 2019-06-17
feat: build compiles down to ES5 syntax

jackmellis:
Currently the build is compiled to ES6-level syntax, which causes issues for anything backwards compatible. In particular webpack + uglify falls over because uglify doesn't understand ES6 syntax. This PR adds some basic `preset-env` to the compilation process.

I also updated to babel 7 for you, hope that's okay!

re: https://github.com/ianstormtaylor/superstruct/issues/83


jackmellis:
bump

jackmellis:
any chance we can move this one forward? :smile:

ianstormtaylor:
Thank you for taking the time to do this @jackmellis, and sorry for the slow response. This looks good to me.

PR #130, opened 2019-06-30
Validate that an interface is an object before checking it's properties

### LINKED ISSUE #119: interface nested in object fail in union
I found a case that I believe should succeed but failed… I haven't had a chance to dig too deep into why yet but I'm filling the issue here to share my finding. :)

I wrote a fixture for testing union of structs:

```js
import { struct } from '../../..'

export const Struct = struct({
  name: 'string',
  age: 'number',
  address: struct.union([
    // This will fail…
    struct.object({
      foo: struct.interface({
        street: 'string',
        city: 'string',
        state: 'string',
        zipcode: 'string',
        country: struct.literal('United States of America'),
      }),
    }),
    /*
    // This would work…
    struct.interface({
      street: 'string',
      city: 'string',
      state: 'string',
      zipcode: 'string',
      country: struct.literal('United States of America'),
    }),
    */
    // If this is place before the first schema, it will work as well…
    struct.object({
      street: 'string',
      city: 'string',
      postcode: 'string',
      country: struct.literal('Singapore'),
    }),
  ]),
})

export const data = {
  name: 'john',
  age: 42,
  address: {
    street: 'Woodlands Street 42',
    city: 'Woodlands',
    postcode: '123456',
    country: 'Singapore',
  },
}

export const output = {
  name: 'john',
  age: 42,
  address: {
    street: 'Woodlands Street 42',
    city: 'Woodlands',
    postcode: '123456',
    country: 'Singapore',
  },
}
```

This will fail with the following error:

```
     TypeError: Cannot read property 'street' of undefined
      at Kind.validate (lib/index.js:543:15)
      at keys.forEach.key (lib/index.js:752:27)
      at Set.forEach (<anonymous>)
      at Kind.validate (lib/index.js:737:10)
      at Kind.validate (lib/index.js:1011:24)
      at keys.forEach.key (lib/index.js:752:27)
      at Set.forEach (<anonymous>)
      at Kind.validate (lib/index.js:737:10)
      at Function.Struct.assert.value [as assert] (lib/index.js:1178:36)
      at Struct (lib/index.js:1165:21)
      at Context.run (test/index.js:32:28)
```

It seems like having a schema with interface nested in object, nested in a union (like above) and it definition come before others schema to be union'ed, will cause the above failure… Note the ordering of schema definition is important…

```js
struct.union([
  struct.object({
    key: struct.interface({
      ...
    })
  }),
  ...
])
```

jormaechea:
Added a validation to check if a `interface` is an object before checking it's properties.
This validation was taken from `partial`.

Test fixture added.
One fixture had to be changed because it's expected error was wrong.

Fixes #119

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -310,6 +310,7 @@ function inter(schema, defaults, options) {
     }
   }
 
+  const obj = scalar('object', undefined, options)
   const ks = []
   const properties = {}
 
@@ -323,6 +324,13 @@ function inter(schema, defaults, options) {
   const name = 'interface'
   const type = `{${ks.join()}}`
   const validate = (value = resolveDefaults(defaults)) => {
+    const [error] = obj.validate(value)
+
+    if (error) {
+      error.type = type
+      return [error]
+    }
+
     const errors = []
     const ret = value
 
--- a/test/fixtures/interface/invalid-nested.js
+++ b/test/fixtures/interface/invalid-nested.js
@@ -1,0 +1,20 @@
+import { struct } from '../../..'
+
+export const Struct = struct({
+  id: 'number',
+  person: struct.interface({
+    name: 'string',
+    age: 'number',
+  }),
+})
+
+export const data = {
+  id: 1,
+}
+
+export const error = {
+  path: ['person'],
+  value: undefined,
+  type: '{name,age}',
+  reason: null,
+}
--- a/test/fixtures/interface/invalid.js
+++ b/test/fixtures/interface/invalid.js
@@ -8,8 +8,8 @@ export const Struct = struct.interface({
 export const data = 'invalid'
 
 export const error = {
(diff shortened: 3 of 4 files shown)


ianstormtaylor:
Hey @jormaechea thank you! Can you add a test that ensures that `interface` still works with an instantiated class and not just plain objects. Just for checking us going forward.

jormaechea:
Sure! I'll try to get this done today or maybe tomorrow. :+1:

jormaechea:
Ready to go! :smile:

ianstormtaylor:
Beautiful, thank you @jormaechea!

jormaechea:
@ianstormtaylor when do you think that you can release this to npm?
Also, there is PR #133 that could be released too if you approve it.. :)

PR #132, opened 2019-07-01
Fixed default values defined as functions for interface structs

### LINKED ISSUE #131: Error with interfaces default values
There is an error when setting a default value as a function in interfaces.

## Minimal example
```js
'use strict';

const { struct } = require('superstruct');

const ExampleStruct = struct.interface(
  {
    name: 'string',
  },
  {
    name: () => 'John Doe',
  }
);

ExampleStruct({ age: 26 });
```

## Current behaviour
It throws an error: 
```
TypeError: Expected a value of type `string` for `name` but received `undefined`.
```

## Expected behaviour
It should be valid data and add the default `name`:
```js
{
  name: 'John Doe',
  age: 26,
}
```

jormaechea:
Applied a fix based on the `partial` struct.
Added the test fixture to cover this case.

Fixes #131

### DIFF
--- a/src/kinds.js
+++ b/src/kinds.js
@@ -328,9 +328,7 @@ function inter(schema, defaults, options) {
 
   const name = 'interface'
   const type = `{${ks.join()}}`
-  const validate = value => {
-    const resolved = resolveDefaults(defaults)
-    value = resolved ? { ...resolved, ...value } : value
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks @jormaechea!

PR #133, opened 2019-07-07
Not empty list struct

### LINKED ISSUE #129: How to validate a not-empty list?
I need to validate a field that should be a not empty array of some custom type. Example:
```js
const User = struct({
	firstname: 'string',
	lastname: 'string'
});

const Struct = struct({
	myProp: struct.list([User])
});

const validData = {
	myProp: [{
		firstname: 'John',
		lastname: 'Doe'
	}]
};

const invalidData = {
	myProp: []
};

// This is fine
console.log(Struct(validData));

// This should fail
console.log(Struct(invalidData));
```

Is there a native way to do this?

jormaechea:
Added a new struct `notEmptyList`
Added it's test fixtures
Added it's documentation

It basically works like a `list` but also checks for at least one element in the array.

Fixes #129

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -15,6 +15,7 @@
   * [`intersection`](#intersection)
   * [`lazy`](#lazy)
   * [`list`](#list)
+  * [`notEmptyList`](#notEmptyList)
   * [`literal`](#literal)
   * [`object`](#object)
   * [`optional`](#optional)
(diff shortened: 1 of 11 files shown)


ianstormtaylor:
Hey @jormaechea, I'm sorry for the slow reply on this. I was thinking... this might be better implemented as an option in the 3rd argument to `.list()`, like:

```js
struct.list([User], undefined, { empty: false })
```

That way we don't have to introduce extra struct types in the future for things like this that can be flags instead. Are you down with that?

jormaechea:
Hi @ianstormtaylor. We have mostly opted out of superstruct a while ago already.. But yeah, I might update the PR to use the third parameter :+1: 

Btw, I'd suggest a major refactor of superstruct, to separate each struct in a different file.. :smile:

ianstormtaylor:
@jormaechea sounds good, I'll close this PR until then in case that. Feel free to open another with the change if you end up getting the time. Cheers!

PR #137, opened 2019-09-21
Introduce a custom error object option

aldipower:
Hi!

This introduces an `error` option on the `superstruct` export to define a custom error object superstruct will throw on invalidates.

```js
import { superstruct } from 'superstruct'

import CustomError from './somewhere/CustomError'

const struct = superstruct({
  error: CustomError,
})
```

I need to be able to get rid of the value in the error and I have two important reason for this.

1. I do not want to expose sensitive user data to my logs on my production system.
2. I implemented a length check to a string type with superstruct to prevent database denial of service. But still the DOS is possible by pollute the logs with huge strings exposed in the error messages.

Hope you get the idea and you like it. :)
Have a nice day.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -104,7 +104,7 @@ const user = User(data)
 
 Superstruct supports more complex use cases too like defining list or scalar structs, applying default values, composing structs inside each other, returning errors instead of throwing them, etc. For more information read the full [Documentation](#documentation).
 
-You could also define a custom error object derived from a TypeError by using the `superstruct` export. Maybe you do not want to expose sensitive user data in the logs for example.
+You could also define a [custom error](./examples/custom-errors.js) object derived from a TypeError by using the `superstruct` factory. Maybe you do not want to expose sensitive user data in the logs for example. 
 
 ```js
 import { superstruct } from 'superstruct'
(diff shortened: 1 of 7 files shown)


ianstormtaylor:
Hey @aldipower, thanks for investigating this. It sounds to me, since it's a logging-specific need, that this would be better solved by having some logic that is logging-specific as a wrapping layer around the errors, than by allowing the error to be overridden.

aldipower:
Hey @ianstormtaylor, I thought about this, but this would introduce extra wrapping and differ logic in a lot of places not belonging to superstruct, just to get rid of something that is provided by superstruct. I also have other validation errors that are not thrown by superstruct. So the easiest and most concise way for me is to just log the error collect from the `try`-block without handling extra things and not mix the concerns. 
And it's actually against of one of your own principles too:
`Many don't actually throw the errors, forcing you to wrap everywhere.`
I am would basically end up with the same problem then.
It's not a breaking change, you could still wrap the errors.
What do you think?

ianstormtaylor:
> Many don't actually throw the errors, forcing you to wrap everywhere.

Sorry, that should actually say "branch" instead of wrap I guess. It's talking about a different phenomenon of flow control.

--- 

But I'm down! The one change I'd ask is if you can remove the `static format(attrs)` method (in both the core library and in the example documentation), to make it more clear for people that the only API they need to implement in their errors is the `constructor(attrs)` signature.

aldipower:
Ah ok, to "branch" makes more sense. :)
Cool! 
So I cleaned up the StructError and removed the static formater, also I moved the `examples/custom-errors.js` to `examples/handling-errors.js` and added a new `custom-errors.js` describing the usage of `{error: CustomError}`.
Furthermore I updated the docs.

---
I absolutely have no idea why `examples/custom-errors.js` is failing during linting and the other examples don't. The linter gives me no useful information. Could you help me?

Issue #135, opened 2019-08-27
rename `dict` to `record`?


ianstormtaylor:
Just to line up better with TypeScript's [`Record`](https://levelup.gitconnected.com/advanced-typescript-types-with-examples-1d144e4eda9e).

Issue #123, opened 2019-01-14
How to validate dependent fields?


iaroslavshvets:
Hi! Let's say I have simple example like this:
`y` type is depending on `x` field value, while `x` field has its own validation. How can i access current field values in runtime, to check dependent fields? Usually it's done via passing 2nd argument to validation function like `(value, data) => RESULT`, where `data` contains all values.
How this can be achieved with superstruct? 

NOTE TO EXAMPLE: there is no `data` argument passed to `y` validating function. This won't work.

```
const struct = superstruct({
  types: {
    x: (v) => v === true,
    y: (v, data) => {
       return data.field_1 === true && v === 'string';
    }
  }
});


const Validator = struct({
  field_1: 'x',
  field_2: 'y'
});

const data = {
  field_1:  true,
  field_2:  'string'
};

```

I

drenther:
Use the [`struct.dynamic`](https://github.com/ianstormtaylor/superstruct/blob/master/docs/reference.md#dynamic) API.

Example from the documentation

```js
const map = {
  TEXT: struct({
    content: 'string'
  }),
  IMAGE: struct({
    url: 'string'
  })
}

const Node = struct({
  type: struct.enum(['TEXT', 'IMAGE']),
  value: struct.dynamic((value, parent) => {
    // Make sure your function never comes back with anything but a struct.
    return map[parent.type] || struct('undefined')
  })
})

const Struct = struct({
  nodes: [Node]
})
```

ericdmoore:
Hey @drenther How would you condition the schema based on a nested attribute? I was thinking about this today:

What I want is to make `lang` conditional on the value of `country`

```js
const semiUsefulNestedSchema = struct({
  a: 'string?',
  b: 'string?',
  c: struct.object({
    e: 'string?',
    f: 'string?',
    g: struct.object({
      country: 'string',
      lang: struct.dynamic((value, parent) => {
        return parent.country === 'USA'
          ? struct.enum(['en'])
          : struct.enum(['es', 'fr', 'zn', 'de', 'nl'])
      })
    })
  })
})
```

The following example **does not work** but is **merely an example idea** that maintains backward compatibility, and allow for the authors of dynamic-schema-functions to have a more ergonomic API that permits `absolute addressing` of attributes to use when building conditional schemas. 

[...]

Issue #117, opened 2018-09-29
Validating objects with unspecified properties


doomsower:
Let's say I have a struct: `struct.object({foo: 'string'})`  
Then I pass an object: `s.validate({ foo: 'bar', bar: 'foo' })`  
Currently I am receiving an error with path `[ 'bar' ]` and message `Expected a value of type `undefined` for `bar` but received `"foo"`.`

I think that this behavior is incorrect, and the `path` should be empty array, i.e. the error belongs to whole struct and not to unspecified property. And the message should be adjusted accordingly.

The following use case illustrates why I think so. I face this situation quite often.  There is a form that receives data from backend, user edits the data, then submits it back. The thing is that what backend sends, what backend expects to receive, and what the form internally uses sometimes have different shapes (the form needs some extra data to look nice, or rich text needs to be transformed from markdown to internal representation, etc...). This leads to unspecified properties, and results in validation error. UI then checks paths of errors and displays them on corresponding UI controls. Those errors with empty path array (like multi-field errors) are then displayed in an ugly red box for whole form errors. The problem with current error paths is that there is no way I can know that unspecified fields errors belong to whole form, they look like they belong to particular UI control, but there is no such control on the form.

doomsower:
It seems that it's a one line fix [here](https://github.com/ianstormtaylor/superstruct/blob/5d537e1017dc27faed6ecd9fbb42aa101850d0f0/src/kinds.js#L550), but I understand that it's a) breaking and b) opinionated change.

doowb:
I think you can use [interface](https://github.com/ianstormtaylor/superstruct/blob/master/docs/reference.md#interface) to achieve your use case. This will only assert the properties in the interface, while [partial](https://github.com/ianstormtaylor/superstruct/blob/master/docs/reference.md#partial) will do the same, but only return the properties in the partial.

doomsower:
@doowb ,  `interface` and `partial` do not produce errors on unspecified fields. And I need errors, not sanitized objects. Just double-checked it in node console.

<details><summary>node REPL</summary>
<p>

```js
> const { struct } = require('superstruct')
undefined
> const o = { foo: 'bar', bar: 'foo' }
undefined
> const schema = { foo: 'string' }
undefined
> const inter = struct.interface(schema)
undefined
> const part = struct.partial(schema)
undefined
> const obj = struct.object(schema)
undefined
> inter.validate(o)
[ undefined, { foo: 'bar', bar: 'foo' } ]
[...]

PR #139, opened 2019-10-07
Add typescript

ianstormtaylor:
###### BREAKING

**Several structs have been renamed!** Superstruct tries to mimic established naming schemes whenever possible for its API, and TypeScript is one of our main comparisons. To make things easier for people, we've renamed a few structs to more closely match their TypeScript counterparts:

- The `list` struct is now called `array`.
- The `partial` struct is now called `pick`.
- The `dict` struct is now called `record`.

Hopefully this will make them easier to understand at a glance!

**The `enums` struct has been removed!** This was special-cased in the API previously, but you can get the exact same behavior by creating an using the `array` and `enum` structs:

```js
struct.array(struct.enum(['red', 'blue', 'green']))
```

**The `interface` struct now returns the original, unaltered value!** In an effort to make things more familiar, the `interface` struct now always returns the object that it is called with when it passes validation. So if the object was a function, a function will be returned. This makes it match more closely with the idea of "structural typing" that TypeScript and other typing systems are based on. _If you want the old behavior, use the `pick` struct.

**Computed values function signatures have changed!** Previously a computed value would be called with a signature of `(value, root)` in some cases and `(value, parent)` in others. This was confusing, and the cause for the inconsistency was complex. This logic has been simplified, and now computed values are called with `(value, branch, path)` in all cases.

```js
struct.dynamic((value, branch, path) => {
  value === branch[branch.length - 1] // you can get the value...
  const parent = branch[branch.length - 2] // ...and the parent...
  const key = path[path.length - 1] // ...and the key...
  value === parent[key]
  const root = branch[0] // ...and the root!
})
```

The `path` is an array of keys representing the nested value's location in the root value. And the `branch` is an array of all of the sub values along the path to get to the current one. This allows you to always be able to receive both the **parent** and the **root** values from any location—as well as any value in between.

**The `error.errors` property has been renamed `error.failures`, and isn't cyclical.** It being cyclical caused lots of issues whenever an `StructError` object was attempted to be serialized. And the `errors` property was slightly confusing because the elements of the array weren't full error objects. The new structure is easier to understand and work with.

**The `error.reason` property is no longer special-cased.** Previously you could return a "reason" string from validator functions and it would be added to error objects. However, now you must return an error properties object (with a `reason` property if you'd like), and all of the properties will be added to the error object. This makes Superstruct even more flexible as far as custom error details go.

**The `type` property of structs have been rewritten to be more clear.** This is an implementation mostly, but the `struct.type` string which shows up in error messages have been tweaked to be slightly more clear exactly what type they are checking for.

###### NEW

**Superstruct is now written in TypeScript.** It was rewritten from the ground up to make use of types, and to have better inline documented if you use a TypeScript-compatible IDE. There are probably improvements that can be made, so if you'd like to contribute please do!

**A new `partial` struct mimics TypeScript's `Partial` utility.** The new struct validates that its input partially matches an object defined as a set of properties with associated types. All of the properties of the object are optional.

**A new `size` struct allows validating array and string lengths.** The new struct validates that its input has a certain size, by checking its `length` property. This works strings or arrays.

**You can now provide a custom `Error` setting.** By passing in your own constructor when configuring Superstruct you can have complete control over the exact errors that are generated by structs that fail validation.

---

Fixes https://github.com/ianstormtaylor/superstruct/issues/36
Fixes https://github.com/ianstormtaylor/superstruct/issues/40
Fixes https://github.com/ianstormtaylor/superstruct/issues/62
Fixes https://github.com/ianstormtaylor/superstruct/issues/92
Fixes https://github.com/ianstormtaylor/superstruct/issues/100
Fixes https://github.com/ianstormtaylor/superstruct/issues/98
Fixes https://github.com/ianstormtaylor/superstruct/issues/107
Fixes https://github.com/ianstormtaylor/superstruct/issues/109
Fixes https://github.com/ianstormtaylor/superstruct/issues/114
Fixes https://github.com/ianstormtaylor/superstruct/issues/117
Fixes https://github.com/ianstormtaylor/superstruct/issues/83
Fixes https://github.com/ianstormtaylor/superstruct/issues/123
Fixes https://github.com/ianstormtaylor/superstruct/issues/135
Fixes https://github.com/ianstormtaylor/superstruct/issues/129
Fixes https://github.com/ianstormtaylor/superstruct/pull/137
Fixes https://github.com/ianstormtaylor/superstruct/pull/122
Fixes https://github.com/ianstormtaylor/superstruct/pull/75


DragonRaider5:
@ianstormtaylor as I can't find the commits addressing it, I'll have to ask directly: Which way is one now supposed to throw errors from custom validators? I can't seem to find any examples and just returning any object with some keys/values does not seem to change anything about the error I receive.

PR #290, opened 2020-03-13
Iterate objects with in, not of

charles-toller:
struct.partial is evaluating whether a set of properties are correct for an object. [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) states to iterate over keys in an object, you need for...in loops, not for...of.


artyil:
any update on that one?

jonavila:
@ianstormtaylor Any chance you can review this and release a version with a fix?

PR #374, opened 2020-06-09
fix: partial logic error

hzzcc:


hzzcc:
My mistake, I thought only fields in the patial should be check. 
https://github.com/ianstormtaylor/superstruct/blob/master/docs/reference.md#partial.

PR #390, opened 2020-06-20
Optional properties

thesunny:
The purpose of the PR is to make `object()` types with `optional()` values appear as optional properties on an object. e.g.

```ts
// given
const struct = object({ a: optional(string()) })

// it should return this type (the `?` being the key change)
type X = {
  a?: string | undefined
}
```

There is also a related change to `package.json` which adds a script `watch:types`.  The purpose of this script is so that when there are changes to code, the TypeScript type definitions get updated. During the normal `watch`, type changes are not reflected. A better solution may be to add `build:types` as part of the `watch` script instead of in a separate `watch:types` script.

Stylistically, the type helpers are near the `object` code. There aren't any existing type helpers (at least in the code I saw) so the choice to locate the type helpers near the code is arbitrary. It may be preferable to place them in `utils`; however, the code isn't used anywhere else (ie. it's not shared) so I opted to put it right above the `object` code. Let me know if you prefer it elsewhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "test": "yarn build:types && yarn test:types && yarn build:cjs && yarn test:mocha",
     "test:mocha": "mocha --require ./test/register.cjs --require source-map-support/register ./test/index.ts",
     "test:types": "tsc --noEmit && tsc --project ./test/tsconfig.json --noEmit",
-    "watch": "yarn build:cjs --watch"
+    "watch": "yarn build:cjs --watch",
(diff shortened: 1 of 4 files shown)


thesunny:
@ianstormtaylor 

This passes all tests on my computer but ci is failing. It's unclear to me why this is happening.

ianstormtaylor:
This is amazing, thank you @thesunny! I'll try restarting the CI to see if that fixes it.

Issue #377, opened 2020-06-09
Consider adding example with import *


thesunny:
I mentioned that I had a validation library with the same semantics as this new superstruct.

The one difference was that every method hung off an object which I called `be`.

One benefit was that you only imported one variable. For example:

```ts
import be from "~/lib/be"

const User = be.object({
  name: be.string(),
  age: be.optional(be.number()),
})
```

You can get the same benefits to superstruct with:

```ts
import * as be from "superstruct"

const User = be.object({
  name: be.string(),
  age: be.optional(be.number()),
})
```

I wonder if it might be worthwhile to add this as an alternate usage style to the documentation for users that find importing each function separately is a point of friction.

ianstormtaylor:
That sounds good to me, can go in the guide part of the docs right at the beginning where it talks about importing I think?

PR #391, opened 2020-06-20
Added import using wildcard to docs

thesunny:
As per https://github.com/ianstormtaylor/superstruct/issues/377

With respect to the documentation, I import the wildcard as `be` which is what my library was called but the reason is that it is fairly descriptive. e.g. `be.optional(be.string())`. It also doesn't conflict with other methods like `is`.

However, that's a personal choice. I can change it to something else like `ss` so we'd end up with `ss.optional(ss.string())`.


ianstormtaylor:
Thank you @thesunny! I just changed it to `s` for now so that the docs aren't too opinionated on it, but you're right that folks can choose whatever name like best.

PR #392, opened 2020-06-20
Add `nullable` method to docs

thesunny:
Add `nullable` method to docs.


thesunny:
@ianstormtaylor Sorry, I accidentally merged `test:types` into my `master` so every PR has `package.json` added to it.

After you merge though, it should all be hunky dory I expect.

ianstormtaylor:
Thank you @thesunny!

PR #425, opened 2020-07-10
docs: escape pipe character to fix error properties table

kimroen:
Escapes the pipe character in the `Array<string | number>` type of the Error Properties table. This pipe was splitting the type into two columns, which meant it looked weird, the other columns were shifted, and the description was invisible.

### Before

split column

### After

joined column


kimroen:
I see the document modified here is split up in https://github.com/ianstormtaylor/superstruct/pull/383, but the error fixed here is still present in that version. Just as a heads up!

Feel free to fix it there instead, I don't mind 😄

zenflow:
Note that this will also have merge conflicts with #440. Great to have this PR open though, as a reminder to fix that problem.

ianstormtaylor:
Thank you for this @kimroen! Sorry for the delay.

PR #432, opened 2020-07-15
Update reference docs on `array`

ryami333:
- Removes erroneous semicolon
- Moves two examples to separate lines
- Uses `id: number() -> { id: 1 }` instead of `id: string() -> id: '1'` because I felt that comprehension required _slightly_ more mental gymnastics when the id string happened to also be numerical.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -104,11 +104,12 @@ null
 
 ```ts
 array(number())
-array(object({ id: string() }))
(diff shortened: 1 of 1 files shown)


ryami333:
@ianstormtaylor ?

ianstormtaylor:
Thanks @ryami333!

PR #440, opened 2020-07-27
Fix ability to get multiple failures (#394)

### LINKED ISSUE #394: StructError failures() iterable only returns first error
When using `.failures()` to iterate over all errors (using the latest release, v0.10.11. Node v12.16.1) it only returns the first error.

I'd expect 2 errors to be logged below...

```javascript
const Article = object({
  id: number(),
  title: number()
})

const data = {
  id: 'a string',
  title: 'also a string'
}

const [error, article] = validate(data, Article);

for(const failure of error.failures()) {
  console.log(failure);
}

// Output
// {
//   value: 'a string',
//   type: 'number',
//   path: [ 'id' ],
//   branch: [ { id: 'a string', title: 'also a string' }, 'a string' ]
// }
```

zenflow:
Fixes #394 

## Problem

So I figured it out. Here's the problematic code:
https://github.com/ianstormtaylor/superstruct/blob/b0ca695995b96e4eb4caf0dec09f712654f1d0e1/src/struct.ts#L166-L167
and
https://github.com/ianstormtaylor/superstruct/blob/b0ca695995b96e4eb4caf0dec09f712654f1d0e1/src/struct.ts#L203-L204

In the first snippet, `check` is a `GeneratorFunction` so the return value is always a `Generator` (despite being typed as a simple `Iterator`).

In the second snippet, `toFailures` *sometimes* returns a `Generator`. 

A `Generator` is indeed technically an `Iterable`, but it is more specifically a `IterableIterator` (special kind of `Iterator` which also implements `Iterable`). 

More importantly, a generator's `Symbol.iterator` method should be called **only once**. If it is called a second time, it will return an iterator that completes without yielding any values (even if the first iterator did not consume all the generator's values). See example below for proof.

<details>
<summary>example</summary>

```js
function * generatorFunction () {
  for (let i = 0; i < 10; i++) {
    yield i
  }
}

let generator = generatorFunction()

let [a, b] = generator
let [c, d] = generator

console.log({a, b, c, d})
// { a: 0, b: 1, c: undefined, d: undefined }
```

</details>

So in the second line of each of the above snippets, when we implicitly call `iterable[Symbol.iterator]()` and save only the first value from the resulting iterator, we are losing the rest of the values forever. 

## Solution

Do not attempt to iterate on `Generator`s more than once (b8b6b5758713a39681a69371217464cfa3f9d8b2)

## Additional changes

- Updates to tests
    - Test for multiple failures (with minimal changes to tests) (1ffa1fd7ef772a827989de8299d7bdb924690f8f)
    - Clean up tests: Have all fixtures export `failures` instead of `error` (a8024a4089acfd4d9ef8f635366110d734402bcf, last commit)
    We can revert this commit and clean [this hack](https://github.com/ianstormtaylor/superstruct/blob/1ffa1fd7ef772a827989de8299d7bdb924690f8f/test/index.ts#L23-L27) (committed in 1ffa1fd7ef772a827989de8299d7bdb924690f8f) another way if you prefer.
- Refined some types in the public API (non-breaking changes) (a85c0db51b56af6fb4bc79996c76666dc033a671)
    - Return type of `StructError.failures` is now `Array<StructFailure>` instead of less specific `Iterable<StructFailure>`. 
    This more closely resembles [the current documentation](https://github.com/ianstormtaylor/superstruct/blob/b0ca695995b96e4eb4caf0dec09f712654f1d0e1/docs/reference.md#error-properties) and is the type users probably want.
    - Return type of `StructContext.check` is now `IterableIterator<StructFailure>` instead of less specific `Iterable<StructFailure>`.
    This more accurately reflects the way it has always been.
- Update documentation regarding StructError.failures (d98fe64f3d4e3b170342c65587c8d97c3a15ce75)

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -547,17 +547,17 @@ The error class that Superstruct uses for its validation errors. This is exposed
 
 Each error thrown includes the following properties:
 
-| **Property** | **Type**                 | **Example**             | **Description**                                                                                                                                                                                                        |
-| ------------ | ------------------------ | ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `branch`     | `Array<any>`             | `[{...}, false]`        | An array of the values being validated at every layer. The first element in the array is the root value, and the last element is the current value that failed. This allows you to inspect the entire validation tree. |
-| `path`       | `Array<string | number>` | `['address', 'street']` | The path to the invalid value relative to the root value.                                                                                                                                                              |
-| `value`      | `any`                    | `false`                 | The invalid value.                                                                                                                                                                                                     |
-| `type`       | `string`                 | `'string'`              | The expected type of the invalid value.                                                                                                                                                                                |
-| `failures`   | `Array<StructFailure>`   | `[{...}]`               | All the validation failures that were encountered. The error object always represents the first failure, but you can write more complex logic involving other failures if you need to.                                 |
+| **Property** | **Type**                     | **Example**             | **Description**                                                                                                                                                                                                        |
+| ------------ | ---------------------------- | ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `branch`     | `Array<any>`                 | `[{...}, false]`        | An array of the values being validated at every layer. The first element in the array is the root value, and the last element is the current value that failed. This allows you to inspect the entire validation tree. |
+| `path`       | `Array<string | number>`     | `['address', 'street']` | The path to the invalid value relative to the root value.                                                                                                                                                              |
+| `value`      | `any`                        | `false`                 | The invalid value.                                                                                                                                                                                                     |
+| `type`       | `string`                     | `'string'`              | The expected type of the invalid value.                                                                                                                                                                                |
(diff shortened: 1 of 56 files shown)


zenflow:
/cc @ianstormtaylor This is ready for review. Hoping we can get this merged! 😁 Let me know if you have any issues with these changes and I will make updates accordingly.  

P.S. I recommend reviewing each commit individually. Each commit is pretty small, except for the last one (clean up tests, a8024a4), but that commit is just the exact same change in 49 different modules, so pretty straight-forward.

zenflow:
@ianstormtaylor Sorry, I forgot to thank you for this incredibly awesome contribution to OSS ❤️  and hopefully also for reviewing this PR 😁

jensdev:
@ianstormtaylor thanks for this package :) Any update on when this PR can be merged?

xiaoyu-tamu:
@ianstormtaylor  Any update ?

fubhy:
This also blocks adoption for us and I assume anyone who wants to use superstruct for form validation.

nmklotas:
Any update for this??

gustavorodarte:
Any update for this??

ianstormtaylor:
Thank you @zenflow, this is an absolute amazing quality PR. I really appreciate it. Sorry for the delay!

zenflow:
@ianstormtaylor No problem! Thanks for the big compliment! 😃

Issue #405, opened 2020-06-27
about partial struct


xtrial:
**from reference:**

`partial` structs are similar to `object` structs, but they only require that the specified properties exist, and they don't care about other properties on the object.

---

but when i try this
```ts
test = { a: 1, b: 2, c: 3 }
struct = partial({ a: number(), b: number() })
assert(test, struct)
```
it throws: ```Expected a value of type `never` for `c` but received `3`.```
i'm expecting it to not throw error.
sorry for my bad english

dwoodwardgb:
Here is the offending line, https://github.com/ianstormtaylor/superstruct/blob/master/src/types.ts#L378

It is trying to ensure that unknown keys aren not present, which should not be happening to partials. It looks like it was copied and pasted from the object type.

ryami333:
Initially I came here for the same reason as the OP, but then I went digging to see if I could fix it and raise a PR and ended up with more questions that answers. There's actually a unit test which explicitly asserts that this current behaviour is the correct behaviour, but as the OP has pointed out - the documentation states the opposite. In the end I was unsure whether in fact it was the documentation which was incorrect, rather than the implementation.

dwoodwardgb:
@ryami333 given the changes in this upcoming breaking PR, I think both the unit test and the code are wrong. The documentation is correct:

https://github.com/ianstormtaylor/superstruct/pull/383

If you check the diff it shows that "partial" has been renamed to "shape" and that it behaves as expected. Not sure when it will be merged.

albertgronholm:
So "partial" just simply doesn't work at all like it's supposed to? Is there any way I can reach similar behavior by doing something else? Do I just have to wait for a fix? Any timeline for such a fix?

Feverqwe:
Until don't fix, wrote custom type
```js
function partial(Structs) {
  const Struct = s.Struct;
  if (Structs instanceof Struct) {
    Structs = Structs.schema;
  }
  const knowns = Object.keys(Structs);
  return new Struct({
    type: `Partial<{${knowns.join(',')}}>`,
    schema: Structs,
    coercer: s.partial(Structs).coercer,
    *validator(value, ctx) {
      if (typeof value !== 'object' || value == null) {
        yield ctx.fail();
        return;
      }
      const unknowns = new Set(Object.keys(value));
      for (const key of knowns) {
        unknowns.delete(key);
        if (!(key in value)) {
          continue;
        }
        const Value = Structs[key];
        const v = value[key];
[...]

PR #496, opened 2020-10-02
Fix partial reference docs

lynxtaa:
Right now implementation of `partial` contradicts its description in docs. Current behavior:
```ts
assert(
  { a:1 },
  partial({ a: number(), b: number() }),
) // ok
```

See https://github.com/ianstormtaylor/superstruct/issues/405#issuecomment-687128744

Fixes https://github.com/ianstormtaylor/superstruct/issues/405

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -286,12 +286,10 @@ partial({
 ```ts
 {
   a: 1,
-  b: 2,
-  c: 3,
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good catch, thank you @lynxtaa!

PR #534, opened 2020-11-13
Add type deprecated

Max-ebay:
Example usage for any type:

```
object({
  deprecatedKey: deprecated(any(), (value, ctx) => console.warn(`${ctx.path} is deprecated`)),
  otherKey: string()
})
```

If this new type is accepted I will update the documentation in this PR.


test/index.ts
```diff
@@ -64,6 +64,18 @@ describe('superstruct', () => {
       }
     })
   }
+
+  describe('deprecated', () => {
```
Max-ebay, line 68: Is there a better place to put this test? I tried to put it into `test/types/deprecated.ts` but due to the custom test files it wasn't executed.

ianstormtaylor:
Hey @Max-ebay this is a really cool idea! I'm down to have this feature added. 

Could you update it so that `deprecated` has a signature of `deprecated(struct, logger)` so that you could do:

```ts
deprecated(string(), console.warn)
```

That way the deprecated values can still be type checked. And people could always pass `unknown()` if they didn't care anymore.

dko-slapdash:
@Max-ebay ?

Max-ebay:
@dko-slapdash Thank you for reminding me. I adapted `deprecated` to accept a `struct`.

ianstormtaylor:
@Max-ebay sorry for not catching this earlier, but I think there's an issue in that the string is hardcoded. In cases where I've handled deprecation before I usually add a custom string that explains what version it was deprecated in, and/or what new feature to use in its place.

I think instead of:

```ts
deprecated(string(), console.warn)
```

We should have it be structured as:

```ts
deprecated(string(), 'The `user.first_name` property is deprecated, please use `user.full_name` instead.')
```

Or for custom cases:

```ts
deprecated(string(), (value, context) => {
  // custom warning logic here...
})
```

And it would default to `console.warn`, but able to be passed as a function instead to completely define your own warning logic (eg. if you want to ensure it only warns once).

Does that make sense?

Max-ebay:
@ianstormtaylor It makes sense! Thank you for your thoughts. I adapted the code to:

```js
deprecated(string(), (value, context) => {
  // custom warning logic here...
})
```

I'd like to avoid the dependency to `console`. Therefore I didn't add that default method. The user should decide what's the best for the use case. Maybe remote logging or a depreciation counter would fit better.

I also think that the passed `log` function needs to be required, because otherwise the user could use `partial()`.

ianstormtaylor:
Hey @Max-ebay sorry for the trouble, but this PR is against `master` and the latest is all on `main` now. I was trying to figure out how the tests were passing even though `ctx.check` doesn't exist anymore and finally realized where the mismatch was. Could you update it to be off the latest source? Thanks.

src/structs/utilities.ts
```diff
@@ -165,6 +165,24 @@ export function partial<S extends ObjectSchema>(
...
+/**
+ * Create a new struct based on an existing struct, but the value is allowed to
+ * be `undefined` and it calls `log` in case the value is not `undefined`.
+ */
+
+export function deprecated<T>(
+  struct: Struct<T>,
+  log: (value: unknown, ctx: Context) => void
+): Struct<T> {
+  return define('deprecated', (value, ctx) => {
```
Max-ebay, line 177: @ianstormtaylor I adapted the code and really like the new `Context` type without further types. It makes it a lot easier to understand and to declare that `log` function from outside of the library without knowing the internals.

I asked myself, why `context.{branch,path}` are of type `any` and not `string`? Isn't every key of `object` of type string or do you want to cover the array type as well where they keys are `number`?

Max-ebay:
> Hey @Max-ebay sorry for the trouble, but this PR is against `master` and the latest is all on `main` now.

No worries. That's the problem of PRs which are open for too long. 😉

ianstormtaylor:
Thanks @Max-ebay, great PR!

Issue #538, opened 2020-11-21
No struct function in v0.11.0


allroundexperts:
`struct` function is missing in the latest version.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/superstruct/commit/574c51a3b03449b9995d0ed12b23521725773558

PR #539, opened 2020-11-21
fix(#538): Added missing struct function

allroundexperts:
As a result of new version 0.11.0, the `struct` function was removed. This PR adds it back.


ianstormtaylor:
Hey @allroundexperts thanks for this. The `struct` function was actually renamed to `define`, but I've added a deprecated version of it back into the codebase—you'll want to rename it though. Thanks!

Issue #523, opened 2020-10-29
Use readonly modifier in enums?


matthias-ccri:
Hi,

In the [`enums` function](https://github.com/ianstormtaylor/superstruct/blob/210e8ac43c3eb117ae98bd91e4e2913a2222ddc7/src/types.ts#L83), it could benefit from a readonly modifier like so:
```
enums<T extends string>(values: readonly T[]): Struct<T>
```

Allow me to explain...
in my use case, I have a variable with an array of strings:
```
export const things = ['thing1', 'thing2'] as const;
```
Notice the `as const`, without which the type would be `string[]`, but now it is `readonly ["thing1", "thing2"]`.
I use this to build a string union type:
```
export type SThings = (typeof things)[number];
// type SModelTypes = "thing1" | "thing2"
```
But notice the `readonly` modifier that was added to `typeof things`. This causes issues when I try to plug `things` into `enum`.
Note: I can work around this by writing an alternate form, but that's ugly:
```
export const things = ['thing1' as 'thing1', 'thing2' as 'thing2'];
```
And in any case, it's good that it's readonly. I don't want any code to change this array.

Anyway, with `as const`, here's the error I see when I do `enums(things)`:
```
No overload matches this call.
  Overload 1 of 2, '(values: any[]): Struct<any, any>', gave the following error.
    Argument of type 'readonly ["thing1", "thing2"]' is not assignable to parameter of type 'any[]'.
      The type 'readonly ["thing1", "thing2"]' is 'readonly' and cannot be assigned to the mutable type 'any[]'.
  Overload 2 of 2, '(values: any[]): Struct<any, any>', gave the following error.
    Argument of type 'readonly ["thing1", "thing2"]' is not assignable to parameter of type 'any[]'.
      The type 'readonly ["thing1", "thing2"]' is 'readonly' and cannot be assigned to the mutable type 'any[]'.ts(2769)
```
So I guess `any[]` isn't permissive enough!

So hopefully you understand the use case.

Adding `readonly` to the enums parameter doesn't seem to restrict the usage to readonly arrays — you can still pass mutable arrays just fine. It appears to just restrict the enum code from modifying the array.

Note: I am able to work around this by type-casting with the `Mutable` type from the type-fest package.
```
enums(things as Mutable<typeof things>)
```
But of course, with readonly params this wouldn't be necessary, and there are no downsides that I can see.

PR welcome?

ianstormtaylor:
[...]

PR #541, opened 2020-11-21
Use `readonly` modifier in enums to accept readonly const arrays

matthias-ccri:
See https://github.com/ianstormtaylor/superstruct/issues/523


ianstormtaylor:
Thanks @matthias-ccri!

Issue #404, opened 2020-06-26
React native package `main` field related error.


cmnstmntmn:
Hei,

I'm trying to plug supertruct to a react-native app, but i'm getting this error

```
error: Error: While trying to resolve module `superstruct` from file `validation.js`, the package `node_modules/superstruct/package.json` was successfully found. 
However, this package itself specifies a `main` module field that could not be resolved (`node_modules/superstruct/index.cjs`. Indeed, none of these files exist:
```

cmnstmntmn:
any workarounds for this ?

cmnstmntmn:
i found [this](https://forums.expo.io/t/sdk-37-cjs-bundler-error-with-superstruct-library/35630)

roccomuso:
any news on this?

rhythnic:
`package.json` is using `type:module`. For recent versions of Node.js, this means that the entrypoint to the module should be an ECMAScript module; however, `main` is set to a CommonJS file.  This makes Superstruct unusable in any Node.js package of `type:module`.  It's possible to support both ECMAScript modules and CommonJS by using the `exports` key.

```
// package.json
{
  // keep pointing main to cjs for legacy versions of node.js
  "main": "./lib/index.cjs",
  // use conditional exports for recent versions of node.js
  "exports": {
    "import": "./lib/index.es.js",
    "require": "./lib/index.cjs"
  },
  "type": "module"
}
```

isnifer:
@rhythnic why did you mention Node.js? The question was about React-Native. Your PR does not help to fix `superstruct` for RN. @ianstormtaylor too early to close this issue.

isnifer:
But this config for metro bundler helped me:
```js
// https://github.com/ioveracker/demo-react-native-emotion/blob/master/metro.config.js

const { getDefaultConfig } = require('metro-config');

module.exports = (async () => {
    const {
        resolver: { sourceExts },
    } = await getDefaultConfig();
    return {
        resolver: {
            // Add cjs extension so stylis will load.
            sourceExts: [...sourceExts, 'cjs'],
        },
        transformer: {
            getTransformOptions: async () => ({
                transform: {
                    experimentalImportSupport: false,
                    inlineRequires: false,
                },
            }),
        },
    };
})();
```

gabrielgrover:
I am using expo for my react native project and this worked for me

`
const { getDefaultConfig } = require('@expo/metro-config');

const defaultConfig = getDefaultConfig(__dirname);

[...]

PR #566, opened 2020-12-11
Add conditional exports to package.json

rhythnic:
This PR is in reference to my comment on #404 .  In brief, Node.js projects of type `module` are not able to import Superstruct without this change to `package.json`

I tested this locally and was successfully able to import and use Superstruct into a Node.js package of type `module`.


ianstormtaylor:
@rhythnic that is awesome, thank you! Had no idea that feature existed—cool to learn.

PR #595, opened 2021-01-01
Fix README: `struct` => `define`

Flaque:
Uses the non-deprecated version in the getting started example.


ianstormtaylor:
Thank you @Flaque!

Issue #568, opened 2020-12-14
custom error messages


jota:
Hey there,
great work with superstruct. I was especially happy to see that it does not bloat the browser bundle like other validation libs do.

One thing is holding me back to switch to superstruct though,

I could not find an easy way to provide a custom error message.

I was looking for something as easy as
pattern(string(), /^\d\d\d\d-\d\d-\d\d$/, 'custom error message, the string should be in date format YYYY-MM-DD')

or
message(pattern(...), 'custom error message, the string should be in date format YYYY-MM-DD')

The default error message provides the regex used which is not something I would like to bother a user with and there are some other use cases like localization.

Is there a way to do that already?

ianstormtaylor:
Hey @jota can you explain your use case fully? Eg. what the specific struct is used to validate, where these error messages are going, etc. (with code samples)

jota:
Hi Ian,

sure. Basically I would like to override the generated error message.

Let's say I have an object with a "date" field where I validate the format YYYY-MM-DD e.g. 2020-12-01.
**Validation:** date: pattern(string(), /^\d\d\d\d-\d\d-\d\d$/),
**Input:** 2020-12-XX
**Error message:** 'At path: date -- Expected a string matching `/^\\d\\d\\d\\d-\\d\\d-\\d\\d$/` but received \"2020-12-XX\"'

I would like to change the error message to something else like "date should match format YYYY-MM-DD" because i would like to directly show those messages in a user interface.

I have created a jsfiddle [here](https://jsfiddle.net/t1p3mraL/)

I'm currently using yup, where I can use the following:
date: yup.string().required().matches(/^\d\d\d\d-\d\d-\d\d$/**, 'date must have format YYYY-MM-DD'**),

Does that make sense to you?

ianstormtaylor:
My general recommendation is that I wouldn't necessarily show these error messages directly to a user, because Superstruct provides no guarantee that they are user-facing. Depending on the structure of your validation they could be very confusing.

I don't think I want to add a `message` argument to _every_ struct because it gets tedious, but also because it makes it hard to compose them if the messages themselves assume things about where the structs are used.

That said, I'd be open to a `message(..., string)` utility function if you were down to pull request one.

The other solution for you is to define a custom type of your own, especially if you are planning to re-use those date strings in other places in your app, for example:

```ts
import { define } from 'superstruct'

const datestring = () => define('datestring', value => ... || 'My custom message.')
```

[...]

PR #611, opened 2021-01-14
feat(reuired): add required with custom message

EvgenBabenko:
New refinements - required
Contains [custom errors](https://github.com/ianstormtaylor/superstruct/issues/568)

Technically this is the same as using size, but "required" feels slightly nicer and will give a slightly easier to read error.
i borrowed this comment from empty

### DIFF
--- a/src/structs/refinements.ts
+++ b/src/structs/refinements.ts
@@ -27,6 +27,17 @@ export function empty<
   })
 }
 
+export function reqiured<T extends string | any[], S extends any>(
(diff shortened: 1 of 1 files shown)


EvgenBabenko:
@ianstormtaylor Is it useful refinements? if not i'll close PR

ianstormtaylor:
Hey @EvgenBabenko, I'd prefer to go with an alternative solution as being discussed in #568, thanks.

PR #629, opened 2021-01-26
make mask() recursive, allow type() pass unknown properties through in mask=true mode

### LINKED ISSUE #618: mask() is not recursive
https://codesandbox.io/s/superstruct-ff5i4-forked-ye1ns?file=/src/index.js

Notice that `zzz` is silently removed (expected), but `unk` is not (surprising). 

```
import { type, mask, number } from "superstruct";

let data = {
  id: 1,
  zzz: 444,
  some: {
    a: 10,
    unk: 55
  }
};

let Type = type({
  id: number(),
  some: type({
    a: number()
  })
});

console.log(JSON.stringify(mask(data, Type)));

// {"id":1,"some":{"a":10,"unk":55}} 
```

P.S.
I don't want to wrap all the type() I have in the code with masked(), because it produces a lot of boilerplate. I assume that the decision of stripping unknown properties is to be made at the caller side (which is likely singular) and not at the definition side (which has way higher cardinality).

### LINKED ISSUE #619: let type() still keep unknown props when mask() is used
Hi.

Would be cool if type() allowed unknown props even if mask() is used for validation. Then, people would use object() almost everywhere, but in some rare cases they would be able to fallback to type() - and still be happy with mask().

I noticed that type() allows unknown props when used in create(), but it still strips them when used in mask(). And there is no way to tell mask() to "strip all unknown props, but for SOME EXPLICIT places - allow them".

I.e. the current logic is binary: either "throw on unknown props" (object() + create()) or "strip all unknown props (object() + mask() or type() + mask()). And there is nothing in between.

When mask() is used, object() behaves exactly the same way as type(), they're no different. I.e. in the code which uses mask() intensively (and I suspect that it is the absolute most of the code, because otherwise there would be problems with client-server backward compatibility, see below), there is no real reason in having two things - object() and type() - only one thing would be enough. Making type() preserve unknown props even when using with mask() would give its existence in a "mask()-preferred environment" a reason.

P.S.
Why I think mask() is usable in like 99% and create in like 1%: imagine we have an old client which uses `object({ a: string() })` to validate the data came from the server, and the server has just become newer: its own structure is `object({ a: string(), b: string() })`. This is a very common situation. And in this situation, create() is unusable, whilst mask() is usable.

P.P.S. 
Yup's author has recently rewrote yup to TS and made a few architectural mistakes (IMHO). One of the mistakes is to pass not a TS type in generic arguments, but the schema type itself, which blows up IntelliSense and also expands the size of the generated *.d.ts files like 10x. It also slows down the compiler. I don't think yup will survive in that state for long, so superstruct has a great advantage IMHO.

dko-slapdash:
- Make mask() and Struct.mask() running masking recursively, fixes #618
- Let type() still keep unknown props when mask() mode is used, fixes #619
- Add unit tests
- Add a couple more sentences to the docs
- Do not remove masked() coercion for backward compatibility reasons (it may still be useful)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -122,12 +122,27 @@ export function* run<T, S>(
     path?: any[]
     branch?: any[]
     coerce?: boolean
+    mask?: boolean
   } = {}
 ): IterableIterator<[Failure, undefined] | [undefined, T]> {
-  const { path = [], branch = [value], coerce = false } = options
+  const { path = [], branch = [value], coerce = false, mask = false } = options
   const ctx: Context = { path, branch }
 
   if (coerce) {
+    if (
+      mask &&
+      struct.type !== 'type' &&
+      isObject(struct.schema) &&
+      isObject(value) &&
+      !Array.isArray(value)
+    ) {
+      for (const key in value) {
+        if (struct.schema[key] === undefined) {
+          delete value[key]
+        }
+      }
+    }
+
     value = struct.coercer(value, ctx)
   }
 
@@ -143,6 +158,7 @@ export function* run<T, S>(
       path: k === undefined ? path : [...path, k],
       branch: k === undefined ? branch : [...branch, v],
       coerce,
+      mask,
     })
 
     for (const t of ts) {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -129,6 +129,8 @@ export function* run<T, S>(
   const ctx: Context = { path, branch }
 
   if (coerce) {
+    value = struct.coercer(value, ctx)
+
     if (
       mask &&
       struct.type !== 'type' &&
@@ -142,8 +144,6 @@ export function* run<T, S>(
         }
       }
     }
-
-    value = struct.coercer(value, ctx)
   }
 
   let valid = true
--- a/docs/reference/coercions.md
+++ b/docs/reference/coercions.md
@@ -28,7 +28,7 @@ masked(
 )
 ```
 
-`masked` augments an object struct to strip any unknown properties from the input when coercing it.
+`masked` augments an object struct to strip any unknown properties from the input when coercing it. It affects the direct properties of the struct only and not the nested structs.
 
 ### `trimmed`
 
--- a/docs/reference/types.md
+++ b/docs/reference/types.md
@@ -214,7 +214,7 @@ struct({
 
 `object` structs validate that a value is an object and that each of its properties match a specific type as well.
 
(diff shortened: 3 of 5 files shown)


dko-slapdash:
Hope this PR's code quality passes the library's standards bar; if not - please advice what to improve, I'm happy to do it.

src/utils.ts
```diff
@@ -122,12 +122,27 @@ export function* run<T, S>(
...
     coerce?: boolean
+    mask?: boolean
   } = {}
 ): IterableIterator<[Failure, undefined] | [undefined, T]> {
-  const { path = [], branch = [value], coerce = false } = options
+  const { path = [], branch = [value], coerce = false, mask = false } = options
   const ctx: Context = { path, branch }
 
   if (coerce) {
+    if (
```
ianstormtaylor, line 132: I believe the masking should happen after the coercing below, because the coercing is what guarantees we've cloned the input. Otherwise as written this will be mutating the input unexpectedly.

ianstormtaylor (commented):
Looks great, thank you for the clean PR @dko-slapdash. Just one comment inline about a tweak I think we need to make.

src/utils.ts
dko-slapdash: Cool, will do. In the original `masked()` implementation, masked's coercion was running first though, and then the struct's own coercion:

- https://github.com/ianstormtaylor/superstruct/blob/140461d90686d76d27efb938a2de92fdc382fcd7/src/structs/coercions.ts#L25
- https://github.com/ianstormtaylor/superstruct/blob/140461d90686d76d27efb938a2de92fdc382fcd7/src/structs/coercions.ts#L80

dko-slapdash:
I just pushed the commit which flips the order of masking and coercion as you requested. Hope you're fine with the difference in behavior in comparison to the upstream where masking was running vice versa, i.e. before the own coercion. 

I actually think that masking-after-coercion is better, because it gives the custom coercion functions more freedom (e.g. they can do some backward compatibility work with the previous props before those props are wiped).

ianstormtaylor:
@dko-slapdash interesting, yeah I think coercing before masking makes the most sense. 

I think I want to get rid of `masked` anyways, since 90% of the use case I think is in using `mask` to omit things on-demand instead of always masking. So no worries about the slight change in behavior because I'll likely do a breaking change for that already.

dko-slapdash:
Thanks Ian! This PR would finalize our migration process from yup.

ianstormtaylor:
@dko-slapdash that's great to hear!

PR #634, opened 2021-02-01
Typed error

kamijin-fanta:
ref #568

A design proposal for structurally expressing errors. This PR makes it possible to support custom error messages and i18n. This library will be available for displaying errors to users.

For example, `size(string(), 1, 100)` outputs the error message` ${expected} with a length ${of} but received one with a length of ' ${length}'`. This change outputs the following error content in addition to the message. It is possible to format the error message according to the context and language.

```ts
interface SizeErrorDetail<T extends number | Date> extends ErrorDetail {
  class: 'size'
  actually: T
  min: T | undefined
  max: T | undefined
  minExlusive: boolean
  maxExlusive: boolean
}
```

- todos
  - resolve conflicts
  - think about compatibility

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -2,7 +2,7 @@
  * A `StructFailure` represents a single specific failure in validation.
  */
 
-export type Failure = {
+export type Failure<E extends Error> = {
   value: any
   key: any
   type: string
@@ -10,6 +10,37 @@ export type Failure = {
   message: string
   branch: Array<any>
   path: Array<any>
+  detail: E
+  failures?: Failure<E>[]
+}
+
+export type Error = ErrorDetail | never;
+export interface ErrorDetail {
+  class: string;
+  message?: string;
+}
+export interface BasicErrorDetail extends ErrorDetail {
+  class: 'basic';
+  except: string;
+  actually: string;
+}
+
+export interface TypeErrorDetail extends ErrorDetail {
+  class: 'type';
+  except: string;
+  actually: unknown;
+}
+
+export interface ValuesErrorDetail<T> extends ErrorDetail {
+  class: 'values';
+  except: T[];
+  actually: T;
+}
+
+export interface ValueErrorDetail<T> extends ErrorDetail {
+  class: 'value';
+  except: T;
+  actually: unknown;
 }
 
 /**
@@ -21,24 +52,26 @@ export type Failure = {
  * continue validation and receive all the failures in the data.
  */
 
-export class StructError extends TypeError {
+export class StructError<E extends Error> extends TypeError {
   value: any
   key!: any
   type!: string
   refinement!: string | undefined
   path!: Array<any>
   branch!: Array<any>
-  failures: () => Array<Failure>;
+  failures: () => Array<Failure<E>>;
+  detail: E;
   [x: string]: any
 
-  constructor(failure: Failure, failures: () => Generator<Failure>) {
-    let cached: Array<Failure> | undefined
+  constructor(failure: Failure<E>, failures: () => Generator<Failure<E>>) {
+    let cached: Array<Failure<E>> | undefined
     const { message, ...rest } = failure
-    const { path } = failure
+    const { path, detail } = failure
     const msg =
       path.length === 0 ? message : `At path: ${path.join('.')} -- ${message}`
-    super(msg)
-    Object.assign(this, rest)
+      super(msg)
+      Object.assign(this, rest)
+    this.detail = detail;
     this.name = this.constructor.name
     this.failures = () => {
       return (cached ??= [failure, ...failures()])
--- a/src/error.ts
+++ b/src/error.ts
@@ -19,10 +19,9 @@ export interface ErrorDetail {
   class: string
   message?: string
 }
-export interface BasicErrorDetail extends ErrorDetail {
-  class: 'basic'
-  except: string
-  actually: string
+export interface GenericErrorDetail extends ErrorDetail {
+  class: 'generic'
+  message: string
 }
 
 export interface TypeErrorDetail extends ErrorDetail {
--- a/src/error.ts
+++ b/src/error.ts
@@ -14,33 +14,38 @@ export type Failure<E extends Error> = {
   failures?: Failure<E>[]
 }
 
-export type Error = ErrorDetail | never;
+export type Error = ErrorDetail | never
 export interface ErrorDetail {
-  class: string;
-  message?: string;
+  class: string
+  message?: string
 }
 export interface BasicErrorDetail extends ErrorDetail {
-  class: 'basic';
-  except: string;
-  actually: string;
+  class: 'basic'
+  except: string
+  actually: string
 }
 
 export interface TypeErrorDetail extends ErrorDetail {
-  class: 'type';
-  except: string;
-  actually: unknown;
+  class: 'type'
+  except: string
+  actually: unknown
 }
 
 export interface ValuesErrorDetail<T> extends ErrorDetail {
-  class: 'values';
-  except: T[];
-  actually: T;
+  class: 'values'
+  except: T[]
+  actually: T
 }
 
 export interface ValueErrorDetail<T> extends ErrorDetail {
-  class: 'value';
-  except: T;
-  actually: unknown;
+  class: 'value'
+  except: T
+  actually: unknown
+}
+
+export interface ThrowErrorDetail extends ErrorDetail {
+  class: 'throw'
+  error: any
 }
 
 /**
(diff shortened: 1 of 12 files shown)


ianstormtaylor:
@kamijin-fanta this is cool! A few ideas...

- I think the `E` generic should maybe represent just a dictionary of values that will be attached to a struct error instead of needing for it to `extends Error` itself? 
- I think the details should be mixed directly into the error object, instead of living in a `details` property.
- We can make the existing `Failure.refinement` prop part of the error dictionary properties instead, which is nice. And we can do the same for `type` too.
- I don't think the `SizeErrorDetail` should be common across `min/max` like that, and instead each refinement should be self-contained.
- Couldn't tell if this already happens, but it would be good to keep the ability to return `true/false` instead of requiring using a details object, since this makes the library really easy to pick up for people.
- The `InferError` utility is a great idea!

Which would make a `min(number(), 0)` struct look something like:

```ts
Struct<number, null, {
  type: 'number'
} | { 
  refinement: 'min'
  min: 0,
  exclusive: false
}>
```

And `Failure` might look like:

```ts
type Failure<E extends Record<string, any>> = {
  value: any
  key: any
  branch: Array<any>
  path: Array<any>
  message: string
} & E
```

kamijin-fanta:
@ianstormtaylor Thanks for the idea!

The reason I made `min / max` common is that I feel that I often need it when displaying error messages. For example, "A value of 10 to 20 is allowed. It was actually 25."

I work on a Failure type change.

kamijin-fanta:
Do you have any idea for the definition of toFailure, Result and Validator when rewriting to `type Failure E extends Record<string, any>` ? For example, it is difficult to generate `Failure<E>` from the following Result type.

```ts
export type Result<E extends Record<string, any>> =
  | boolean
  | string
  | Partial<Failure<E>>
  | Iterable<boolean | string | Partial<Failure<E>>>
```

It may be possible to solve it by redefining the Result type as follows.

```ts
export type DescribedResult<E extends Record<string, any>> =
  | boolean
  | string
  | Partial<Failure<E>>
  | Iterable<boolean | string | Partial<Failure<E>>>

export type SimpleResult = boolean | string | Iterable<boolean | string>

export function define<T>(
  name: string,
  validator: SimpleValidator
): Struct<T, null, {}>;
```

Also, it can be easily implemented by giving up the type definition for detailed error information. Of course I want to avoid this if possible.

ianstormtaylor:
@kamijin-fanta good question! I tried messing with it in the TypeScript Playground and came up with [this approach](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwDYEMDOa4GUZQK4IwA8AKnALxx4B2A1tRAO7UA02FVdDzbACnKBjBqAE0y8oEMJkoBvAL4A+OLIBQcOFGAoREakgCecEgE1eAUQCEALmPq4MA2GC20uAJbUA5vbQIAFsAAtii2WPYIEMBQCNG2ABQAbihIeC6c9EysiHpCIDC2AMK5ggCUFMo0mcz2yUjuIigw0Al1abZV3NmR1HkFcMW9ZRVwAJJCUCgARkjARABiKO6pWkS8ior2WgBmnnFwSSntxmw9fUUl+eXkyuPR07MLSytz65sawh7AaAn2Gm3pTpZFh-HJDfIXcEwezXW4TB5zUEAbVBGjcUE8XjgAB84NQ8EEptEQRpSRkuiSydhcARiChqAY2ECeHB1jjqfhCERqMBEsTycDWe9SQBdeybCJ6dG06DxMCSaS2NRkxzOVwebygvyBEJhUGRaKxKAAfguhuioLqDSa0FNcAAailrc0oGthRodnsTbYAErAXY810kPjugS9DHfO04TnEYPYENIgDknwjaETYo08nKytJPTcKlRDicwEpZO1wRQpdJVsaLqrHv9XvrOXNUA4h1SgK4WVhcABzZT7m+HG2NEI7j0ACoDtn5FX5Bx5VI0KpQTB-O40AA6VXADi7tcb7flkIcE8oQ+breD4eUG8rlVHrcGmLRDgvo2rsnubYHGs2qBs0Ldcr3-F1237ME+l7HMqTBfMtDQPAkHgSgwNlSCzjKQsNB-A5EOQ1DyEoGk92xXECJQihiM4ERGx5ERyi0GA8Cgag4CRDM4M0YAWLYhwIEWZZWO+eJKJgU5LgkvsjmAUpC3kUEF2AJA0D3WDSRA7d0LbSh4l7TilK-Uk8LE+joiAuCtK3T1Awg2TJKhGCcPg+BxI4Wzog7NJHOglzTPc4iSPwMiKO+QjqMoGg6IDYBGJ4vj2MM7iEtY9jmiEl40DMpCUN8wQ2ABeS4MUsllNU9TgKfTzdJnEZktJUq4EUxTVF3VkFRkOAICmAArYBCFXdrBj6DgcymSZqACWwAEEoEmAwiHpAxhTAJp-Fm+aUEW5bNla4aolfIMOGZahlD0gFbDjLCIQGKTe1OobiwdJ1a2gNYBHyYQxA65dzr-WSOm7Fkbv6EbhhuOA-VymA0DdJ7nCh8ygz4T6hFEcROv+7z0muqTIWgkZocIuH1gRvdMpEj7BG+zG-rG2pAbgZb7FoYADFsFmNCCb40BQLx0nRTF7Am+lprgOaFqWhlVvWzapd21QFwAMlZcmkZh6mvox37pH+-gaZ1hQ4GNOB7FxKYIAgWZ6XNuAhc1DRcX4WxeHV4mUNJtHad1tB-o94g2VxO5JhmOYA7dCVRymmAJ3SiAJGXLX0Z+xO9ZywjfXClC3VKV2CzJZi0oOXcIF-QKaMTS3re0ahExxXFS-L7OiMoRMHa8RMXNN43bHE8p0GZhkldXaPxz0ATKdWA3tdTrH4lBcSs819Yq1BgmCstJnlpBPO4Cn15lFxaLzJEAu4AClvIocELLNJIu2PsVqNDzeAl2kfcE86jOUOK1yVDgGtdcbBRZTX8M1d8UlJTUHzKhOAiZdz108IAzqJs4AAAN2oAB10EABJZBylQYPZapQdzFnkDg9BcBbCJhQPbZwCAfzuAQEWZwXcX5SjchwRMnkea9CQexd+mBTZUMYO4dcPFYp8PgDg-BhDlzM0wCQmy5lpEUPQVQmh7D-48wwPzPclB0HmHAANIQZ86EAm6r+fBMB5D4KgHOOAUw8BuQGsAdwfIRC2FkbIeUngYA41KOo9BWxeLFw0v2UEbMOaAPWkiIB-gtyzG8BIgAtHAAAjCKKsui+YCyrAkqsoCAhVi3GUoRlIWqjzHLHKck9ngiS9obOef0F4Nhhj8DWJM3SUnXndKElJLpD0ZKoPeIcETjJdE8YS08Njn1MpYTc4yw4-1huUcirDgBlx4h06+7cNSdzvu0kmHAkTiTQFxZ+cBtjQAOK-HZEVtnnKOf-bYDStCfwPqs-K+RCqyXkqCUybyZlyXPmSAwQ4kBn2BS8IymYR6qDHrU9iSz4Rh1IIoeIIAgbVGoHvEAF9MDLMeCQZQOYH4orQAAeT6qYrF5RlaqybgcAlxCGSlCRFgAwhJrZbnEfcF0Io9lIrjl3VqIqJ6bhpf1QgWKcVdHxYS7qtLCDnwpZs7ZBKgrwJ6jKmA9dGVwAJZYSg+IkBIARRKwRGJeg405uy9UNqsTkrCfxZlljtX7KdfXU2AApLAVKAByW4O4-gMIE6hGD8EAnkCE8VNS45wBinsUgqNml03TvYagKAeaOsxIM16AFbCOnqG9V0ZM97RlpKmvEyEkAhjVa69iPJGAclpPEWQmzbDZp5mwc83a61-NLQBZq8lWoAHpx1wFSTO1c9y6GUGTTyeItDExsAjZDDSpl3WyT2R3RMLzC5NpvmkJSAgKpgqpOqt5FUjJZjnZwpxHAl3ABXVMNdANOzOW-L+eIO7Ox7oOQey9R7EonuAGelSakQP32PYmXJ+jtHwvvaoe5LDF3mRXQgD9G6yWAt-f+tIgHvWHtg2B0ikGL0aVA+ElBUgaGeBufXJqyH5JAA) which may help? It keeps a "props" dictionary as the third argument which then determines which "results" are allowed to be returned by a validation function.

I wasn't sure how to limit it more than `object`, but it might work for now?

xiaoyu-tamu:
Hi, what's the status for this PR?

kamijin-fanta:
@xiaoyu-tamu sorry. I couldn't get enough time and left it. I would like to work on it when other work is done, but I would like to welcome other people's contributions.

haf:
Maybe the maintainers would consider taking this code in and continuing on it?

PR #639, opened 2021-02-01
fix: fixed describe for some literal types

freshollie:
Small PR to fix some interesting false error cases of using describe with literal types:

```ts
const a: Describe<1 | 2 | 3> = enums([1, 2, 3]) // type error
const b: Describe<'1' | '2' | '3'> = enums(['1', '2', '3']) // works

const c: Describe<'test'> = literal('test') // type error
const d: Describe<1'> = literal(1) // works
```

I've added tests for the failing cases, with the fix being to provide the same schema type as `enum` when using `literal`.

### DIFF
--- a/src/structs/types.ts
+++ b/src/structs/types.ts
@@ -233,16 +233,31 @@ export function intersection(Structs: Array<Struct<any, any>>): any {
  */
 
 export function literal<T extends boolean>(constant: T): Struct<T, null>
-export function literal<T extends number>(constant: T): Struct<T, null>
-export function literal<T extends string>(constant: T): Struct<T, null>
-export function literal<T>(constant: T): Struct<T, null>
+export function literal<T extends number>(
+  constant: T
+): Struct<T, { [K in T[][number]]: K }>
+export function literal<T extends string>(
+  constant: T
+): Struct<T, { [K in T[][number]]: K }>
+export function literal<T>(
+  constant: T
+): Struct<T, T extends string | number ? { [K in T[][number]]: K } : null>
 export function literal<T>(constant: T): any {
   const description = print(constant)
-  return define('literal', (value) => {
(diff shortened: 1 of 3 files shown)


freshollie:
@ianstormtaylor sorry to ping, is this not a fix you are looking for? Is there some underlying problem with how these types are represented which shouldn't be solved this way? Looking for feedback.

src/structs/types.ts
```diff
@@ -233,16 +233,31 @@ export function intersection(Structs: Array<Struct<any, any>>): any {
...
-      `Expected the literal \`${description}\`, but received: ${print(value)}`
-    )
+  return new Struct({
+    type: 'literal',
+    schema:
+      typeof constant === 'string' || typeof constant === 'number'
+        ? {
+            [description]: constant,
+          }
+        : null,
```
ianstormtaylor, line 254: Can you explain how this `schema` property works? It looks strange to me, but I assume it was necessary to do it this way for some reason?

ianstormtaylor (commented):
Hey @freshollie, not at all—I just missed the notification for this one sadly. It looks great actually, just one question inline. Thanks!

src/structs/types.ts
freshollie: I've followed the same convention as `enums` for schema. Effectively this makes `literal('value')` and `enums(['value'])` the same far as types are concerned.

ianstormtaylor: Ahhh I see, because otherwise the `string` types weren't being inferred correctly. Good idea!

I think we might want literals to have a schema of the literal itself directly, instead of nested inside a dictionary. But this likely requires a bunch of weird TypeScript wrangling, so I'll test out if that's possible in another commit after merging. Thank you!

PR #651, opened 2021-02-08
docs: replaced deprecated with current API use in introduction docs

bobbyhadz:
Replaced the use of the deprecated `struct` function in favor of `define`.

Instead of incrementing the not declared variable `i`, in my opinion the example on line 70 gets the point across better if it's self sufficient - you can copy and paste it and it works.


ianstormtaylor:
Thanks @bobbyhadz, great PR!

PR #653, opened 2021-02-14
improve defaulted coercions types

xiaoyu-tamu:
Is there any special reason we use any here? I think the defaulted value should match the struct shape. with this simple change, we could get autocomplete from editor.

image

image

### DIFF
--- a/src/structs/coercions.ts
+++ b/src/structs/coercions.ts
@@ -37,7 +37,7 @@ export function coerce<T, S, C>(
 
 export function defaulted<T, S>(
   struct: Struct<T, S>,
-  fallback: any,
+  fallback: Struct<T, S>['TYPE'],
   options: {
     strict?: boolean
   } = {}


xiaoyu-tamu:
well, it should be `DeepPartial<Struct<T, S>['TYPE']>`.

where
```
type DeepPartial<T> = T extends Array<infer U>
  ? Array<DeepPartial<U>>
  : T extends ReadonlyArray<infer U>
  ? ReadonlyArray<DeepPartial<U>>
  : T extends {
      [key in keyof T]: T[key];
    }
  ? {
      [K in keyof T]?: DeepPartial<T[K]>;
    }
  : T;
```

I will update the type and fix the test if you ok with this change

ianstormtaylor:
@xiaoyu-tamu the issue is that there are use cases where you want to provide a default that doesn't necessarily match the shape of the struct. For example providing `{}` at the parent level with children with their own defaulters defined as well. Or for more complex coercions where you might provide a default that gets coerced again.

xiaoyu-tamu:
That make sense. It could solved in user land if they need strict type.

PR #691, opened 2021-03-02
Readme: `struct` => `define` in usage example

gr2m:
Follow up to https://github.com/ianstormtaylor/superstruct/pull/595. Seems that the `struct` import has been overseen?


ianstormtaylor:
Thanks @gr2m!

PR #721, opened 2021-03-23
feat(assign): return type struct when assigning to type struct

### LINKED ISSUE #720: `assign`ing `type` structs results in `object` structs
```
const Vehicle = type({ speed: number() });
const Bicycle = assign(Vehicle, type({
  gears: number()
});

assert({ speed: 13, gears: 1, wheel: 1 }, Bicycle)
// StructError: At path: wheel -- Expected a value of type `never`, but received: `1`
```

Is there a way around this already? Extending 2 types should return a type. Or at least, there should be a method for extending types.

--- 

Caused by `assign` only returning an `object` struct.
```
export function assign(...Structs: Struct<any>[]): any {
  const schemas = Structs.map((s) => s.schema)
  const schema = Object.assign({}, ...schemas)
  return object(schema)
}
```

Could #567 address this?

nhagen:
adds support for extending type structs in a way that preserves the type behavior of the original
struct. Previous behavior would return an object struct with stricter requirements about unexpected
fields, rather than the loose requirements of the base type struct assigned "into".

BREAKING CHANGE: Behavior of struct returned by assign has changed

Being generous with the "BREAKING CHANGE"--one could argue this is a patch for a bugfix.

fix #720


ianstormtaylor:
Thanks for this @nhagen, I agree that this was a bug in the original behavior, it's unintuitive that assigning would change the structs type.

PR #726, opened 2021-03-28
fix(max): amend error message

lttb:
Thank you a lot for your work!

Just a minor `max` refinement error message fix.


ianstormtaylor:
Thanks @lttb!

PR #744, opened 2021-04-08
fix expected output value

carusog:
Based on the example, expected value for line 44 should be `true`, not `false`.


ianstormtaylor:
Thanks @carusog!

PR #755, opened 2021-04-20
feat: not empty refinement

ajsaraujo:
I believe it makes sense to have this refinement. Most of the times you want to ensure that an array/string/set/map is _not_ empty rather than empty.

### DIFF
--- a/docs/reference/refinements.md
+++ b/docs/reference/refinements.md
@@ -46,6 +46,15 @@ max(number(), 0)
 
 > 🤖 If you need an exclusive maxmimum you can pass `{ exclusive: true }` as the third argument, like `max(number(), 0, { exclusive: true })` for positive numbers.
 
+### `notEmpty`
+
+```
+notEmpty(string())
+notEmpty(array())
+```
+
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Hey @ajsaraujo thank you for this! I agree that it's a useful thing to be in core. Can you change the name of the function to `nonempty`? I'm happy to merge if so.

The tests look good—they're a little hard to grok. They test that the return value (after running an assertion) is the same as the generic value you pass in, so that we're sure the typing is able to narrow properly.

shellscape:
@ianstormtaylor what about a general `not` that negates the given parameter? We could simply leverage `not(empty(string()))`

ianstormtaylor:
@shellscape that sounds interesting! I think it would be for a slightly different use case than this one (because `not(empty(string()))` would match anything that's not an empty string). But it sounds like something that would be useful regardless, so if you'd like to open an issue/PR for it that would be great!

ifiokjr:
I can add this if it's held up for whatever reason?

PR #771, opened 2021-04-28
types.md: Explain how to set the Typescript type when using define()

barbogast:
Took me a while to figure this out, so I thought it might be worthwhile to add it to the docs.


ianstormtaylor:
Good call, thanks @barbogast!

PR #899, opened 2021-08-19
Correctly match boolean literal values

### LINKED ISSUE #754: Describe incorrectly handles 'true' and 'false' types
Hey, Ian!

I found a problem while trying to write a struct for a described exact boolean value:
```typescript
const _true: Describe<true> = literal(true);
```

That's what we got here:
```
Type 'Struct<false, false>' is not assignable to type 'Describe<false>'.
  Types of property 'schema' are incompatible.
    Type 'false' is not assignable to type 'null'.ts(2322)
```

On the other hand, this code is correct:
```typescript
const _42: Describe<42> = literal(42);
```

Describe is using utility type `StructSchema`, so it seems that this type has a bug:
```typescript
let _42Schema: StructSchema<42>;           // 42
let _trueSchema: StructSchema<true>;       // null, but should be true
```

Here is the sandbox: https://codesandbox.io/s/fast-water-t00d7

birtles:
`boolean` is effectively the union type `true | false` but the current `isMatch` implementation will treat overlapping union types as matching.

This patch introduces a more strict type equivalence check based on https://github.com/Microsoft/TypeScript/issues/27024#issuecomment-421529650 and uses it when checking types that extend `boolean` for equivalence to `boolean`.

The test added in this patch fails without the corresponding code changes here.

Fixes #754.


ianstormtaylor:
Thanks @birtles!

PR #966, opened 2021-10-21
Respect refiner in `dynamic` and `lazy`

### LINKED ISSUE #830: Refinements are not working within dynamic util
I believe I've found a bug. Refinements don't seem to work when returned from the dynamic util. See below example, which I would expect to fail - however it passes validation.

```javascript
const schema = object({
  name: string(),
  age: dynamic((value, context) => {
    return size(integer(), 1, 100);
  })
});

const person = struct.validate({name: 'Ross', age: 1000}, schema);
```

Gelio:
This PR fixes the refiners being previously ignored in the structs returned by `lazy` and `dynamic` utilities.

I have added 2 new test cases that prove the refiners work now.

Fixes #830

### DIFF
--- a/src/structs/utilities.ts
+++ b/src/structs/utilities.ts
@@ -149,6 +149,10 @@ export function lazy<T>(fn: () => Struct<T, any>): Struct<T, null> {
       struct ??= fn()
       return struct.coercer(value, ctx)
     },
+    refiner(value, ctx) {
+      struct ??= fn()
+      return struct.refiner(value, ctx)
+    },
   })
 }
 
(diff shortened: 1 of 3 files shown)


sammkj:
Is there anything that I can help to get this PR in?

Gelio:
@sammkj Unfortunately, I don't think so. @ianstormtaylor hasn't been active in this repo (or at all, looking at his GitHub profile) since a few days before I opened this PR. At work, we are using a patched version of `superstruct` with this PR applied. You could do the same.

Gelio:
No worries, thanks for taking a look and merging it!

Issue #829, opened 2021-06-12
omit, pick and partial should return schema's type


rozaxe:
When passing a schema to `omit`, `pick` or `partial`, the resulting schema is always a struct `object.()`

It would be great to return the schema's type so that when passing a `type()`, the resulting schema is also a `type()`.

Example:

```js
const TodoSchema = type({
  id: string(),
  message: string()
})

const CreateTodoSchema = omit(TodoSchema, ['id'])
// CreateTodoSchema's is a struct 'object()' while it would great to keep it as a 'type' because it's input is.
```

I am willing to make the relevant PR if you want.

PR #988, opened 2021-12-21
Use type for omit output if that's the struct's input

IgnusG:
This change basically allows a `type` struct as input to `omit` to be preserved. It's not very useful for actual checking since `type` accepts anything (even the "omitted" key) but it does remove the key from the returned type and preserves the intuitive behavior behind `type`.

I believe this relates to https://github.com/ianstormtaylor/superstruct/issues/829 but it's missing implementations for `pick` and `partial`.

### DIFF
--- a/src/structs/utilities.ts
+++ b/src/structs/utilities.ts
@@ -166,7 +166,10 @@ export function omit<S extends ObjectSchema, K extends keyof S>(
     delete subschema[key]
   }
 
-  return object(subschema as Omit<S, K>)
+  switch (struct.type) {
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks for this @IgnusG! If you'd be down to write a PR for `pick` or `partial` I'd be happy to merge those as well.

jon-certik:
Hi @ianstormtaylor, `pick` and `partial` have been fixed by PR #1149 and #1150. Would really appreciate it if these patches can be merged soon! Otherwise the current loopholes significantly undermine the usability of the library on complex types.

PR #1003, opened 2022-02-04
add support for dynamic with refinement

roothybrid7:
Added the ability to use refiner in dynamic, such as `empty` and custom refinement.


Gelio:
Just FYI this seems to be a duplicate of https://github.com/ianstormtaylor/superstruct/pull/966 which also adds support for refiner in `lazy`

ianstormtaylor:
Thank you for this @roothybrid7! I'm going to close in favor of #966 as @Gelio mentioned.

PR #1038, opened 2022-04-01
Fixes #971 - nested coercion

### LINKED ISSUE #971: Coercions nested inside object doesn't work
Example
```
const CoercedDate = coerce(date(),string(), (dateString) => {
    return new Date(dateString);
});

const UserId = string();

const User = object({
    createdAt: CoercedDate,
    name: string()
});

const facilityValidator = object({
    createdAt: CoercedDate,
    user: union(UserId, User),
});

const facility = {
    createdAt: '2021-10-10T10:10:10.100Z',
    user: {
        createdAt: '2021-11-11T11:11:11.111Z', // This should be coerced, but is not
        name: 'hal',
    },
}

const [error] = validate(facility, facilityValidator, { coerce: true, mask: false });
// "At path: user -- Expected the value to satisfy a union of `string | object`, but received: [object Object]"
```

kmannislands:
Started with a failing test per @yss14's example in the ticket

Tried implementing an entries generator per @ianstormtaylor's suggestion but it didn't seem to do the trick.

Seems @poyangliu was correct: the issue is that the `coerce` bit of context wasn't being propagated in the call to `validate()` in the `run()` method. Added it to the argument to `validate` though I'm not sure this is the correct approach. Perhaps `coerce` and `mask` should be optional members of `Context` in general?

Closes #971

### DIFF
--- a/src/structs/types.ts
+++ b/src/structs/types.ts
@@ -491,15 +491,22 @@ export function union<A extends AnyStruct, B extends AnyStruct[]>(
   Structs: [A, ...B]
 ): Struct<Infer<A> | InferStructTuple<B>[number], null> {
   const description = Structs.map((s) => s.type).join(' | ')
+  const getMatchingStruct = (value: unknown) =>
+    Structs.find((s) => {
+      const [e] = s.validate(value, { coerce: true })
+      return !e
+    }) || unknown()
   return new Struct({
     type: 'union',
     schema: null,
+    *entries(value, ctx) {
+      const S = getMatchingStruct(value)
+      const entiresToYield = [...S.entries(value, ctx)]
+
+      yield* entiresToYield
+    },
     coercer(value, ctx) {
-      const firstMatch =
-        Structs.find((s) => {
-          const [e] = s.validate(value, { coerce: true })
-          return !e
-        }) || unknown()
+      const firstMatch = getMatchingStruct(value)
       return firstMatch.coercer(value, ctx)
     },
     validator(value, ctx) {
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -148,7 +148,12 @@ export function* run<T, S>(
 
   let valid = true
 
-  for (const failure of struct.validator(value, ctx)) {
+  for (const failure of struct.validator(value, {
+    path,
+    branch,
+    coerce,
```
kmannislands, line 154: This was the real fix here

src/structs/types.ts
```diff
@@ -491,15 +491,20 @@ export function union<A extends AnyStruct, B extends AnyStruct[]>(
...
+      const S = getMatchingStruct(value)
+      yield* S.entries(value, ctx)
+    },
     coercer(value, ctx) {
-      const firstMatch =
-        Structs.find((s) => {
-          const [e] = s.validate(value, { coerce: true })
-          return !e
-        }) || unknown()
+      const firstMatch = getMatchingStruct(value)
```
kmannislands, line 507: TODO: should this take `ctx` with a `coerce` value as well and pass it through to `getMatchingStruct(value, ctx)` instead of hard-coding `coerce: true`?

src/utils.ts
ianstormtaylor: @kmannislands I'm not sure I follow. This is passing in these new properties, but where are they being used differently than before? As far as I can see from this PR, only the type signature has been modified?

Issue #979, opened 2021-11-22
Refinement for top level object failed when field has refinement


PoyangLiu:
Using the validators below, the top level `PasswordDoNotMatch` refinement will *not* be called.
On the other hand, if we change to `newPassword: string()` (i.e. no refinement in any of the nested fields), then the `PasswordDoNotMatch` *will* be called.

```
const PasswordValidator = refine(string(), 'MinimumLength', (pw) => pw >= 8);
const ChangePasswordValidator = object({
    newPassword: PasswordValidator,
    confirmPassword: string(),
});
const RefinedChangePasswordValidator = refine(ChangePasswordValidator, 'PasswordDoNotMatch', (values) => {
    return values.newPassword == values.confirmPassword;  // This will not be called
});
```

Gelio:
As far as I understand, this is a general problem with refinements for compound structures that they are not executed if there is **any** failure from an inner field, regardless of whether the problem is a refinement error or a validation error. The issue title is misleading - refinements for objects/arrays work fine when there are no issues inside of those objects/arrays, but are not executed when at least one refinement inside fails.

## Reproduction

In [the superstruct JSFiddle](https://jsfiddle.net/85nse1mk/)  I used the following code to reproduce the problem:

```ts
const { assert, object, number, string, refine} = window.Superstruct

const PasswordValidator = refine(string(), 'MinimumLength', (pw) => pw.length >= 8 ? true : 'required minimum length of 8');
const ChangePasswordValidator = object({
    newPassword: PasswordValidator,
    confirmPassword: string(),
});
const RefinedChangePasswordValidator = refine(ChangePasswordValidator, 'PasswordDoNotMatch', (values) => {
    return values.newPassword === values.confirmPassword ? true : "Passwords do not match";  // This will not be called
});

const data = {
  newPassword: "12345678",
  confirmPassword: '123456789'
}

try {
  assert(data, RefinedChangePasswordValidator)
  log('valid', data)
} catch (e) {
  const { message, value, type, path, branch } = e
  log('invalid', { message, value, type, path, branch })



function log(type, data) {
  document.body.className = type
  document.body.textContent = JSON.stringify(data, null, 2)
}
```

When `newPassword` was 8 characters or more, the refinement for the object run and returned the error correctly

![image](https://user-images.githubusercontent.com/889383/169279210-bc096564-d12a-484a-802b-9f1460a03427.png)

[...]

PR #1068, opened 2022-06-05
Run refiners as long as validators passed

Gelio:
Previously refiners ran only when there were no past failures during
validation of the value itself or its subvalues. This behavior was
causing inferior UX when validating forms with additional validations
defined as refiners. If there was a refinement error in some subfield,
the refiners defined on the parent of that subfield were not executed.
Only after fixing the error in the subfield did the user receive
information about the failure on the parent field.

This commit changes the condition under which refinements are run.
Refinements are executed if there were no previous failures or those
failures came exclusively from other refiners.

The change helps in the form validation scenario because as long as the
form structure is valid (which is checked using `validator`s), failures
from all refiners are collected.

Refiners still run only if the general structure of the value is
correct. They will get passed a structurally-valid value. The only
difference now is that those values were may not have been checked via a
refiner.

If some check is so important that other refiners should not
run if that check fails, that check should be defined as a `validator`,
not a `refiner`. Failing that check will not run other refiners.

Fixes https://github.com/ianstormtaylor/superstruct/issues/979

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -146,10 +146,10 @@ export function* run<T, S>(
     }
   }
 
-  let valid = true
+  let validity: StructValidity = 'valid'
 
   for (const failure of struct.validator(value, ctx)) {
-    valid = false
+    validity = updateStructValidityAfterFailure(validity, failure)
     yield [failure, undefined]
   }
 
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks @Gelio!

Gelio:
Awesome, thanks for merging the PR (and the cleanup :sweat_smile: )

PR #1086, opened 2022-07-08
Rename CommonJS bundle so as to be compatible with React Native

### LINKED ISSUE #1085: Rename files to `cjs.js` for compatibility with React Native
Importing @solana/web3.js in React Native throws the following error.

```
error: Error: While trying to resolve module `superstruct` from file `/Users/sluscher/mobile-wallet-adapter/js/node_modules/@solana/web3.js/lib/index.browser.cjs.js`, the package `/Users/sluscher/mobile-wallet-adapter/js/node_modules/superstruct/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/Users/sluscher/mobile-wallet-adapter/js/node_modules/superstruct/lib/index.cjs`. Indeed, none of these files exist:

  * /Users/sluscher/mobile-wallet-adapter/js/node_modules/superstruct/lib/index.cjs(.native|.android.js|.native.js|.js|.android.json|.native.json|.json|.android.ts|.native.ts|.ts|.android.tsx|.native.tsx|.tsx)
```

You can see that the React Native packager expects a `.js` extension in addition to `.cjs`

If `superstruct` could rename the CommonJS bundles from `.cjs` to `.cjs.js` that would make it Just Work™ for React Native folks!

More details on the problem here, if you're interested: https://github.com/facebook/metro/issues/535

steveluscher:
Fixes #1085. Read more about the problem statement there.

#### Summary of changes:

* Tests now run against the raw Typescript source. This was necessary because Mocha is awful and renaming the `cjs` bundle caused the entire test system to fall apart. The one I put in place is faster, and unlocks watch mode (`yarn test --watch`)
* Renamed CommonJS bundle to `cjs.js`

### DIFF
--- a/config/rollup-cjs.js
+++ b/config/rollup-cjs.js
@@ -1,7 +1,7 @@
 import config from './rollup'
 
 config.output = {
-  file: './lib/index.cjs',
+  file: './lib/index.cjs.js',
   format: 'cjs',
   name: 'Superstruct',
   sourcemap: true,
--- a/package.json
+++ b/package.json
@@ -7,11 +7,11 @@
   "repository": "git://github.com/ianstormtaylor/superstruct.git",
   "source": "./src/index.ts",
   "types": "./lib/index.d.ts",
-  "main": "./lib/index.cjs",
+  "main": "./lib/index.cjs.js",
   "module": "./lib/index.es.js",
   "exports": {
     "import": "./lib/index.es.js",
-    "require": "./lib/index.cjs"
+    "require": "./lib/index.cjs.js"
   },
   "sideEffects": false,
   "files": [
--- a/package.json
+++ b/package.json
@@ -28,12 +28,14 @@
     "@babel/preset-env": "^7.6.3",
     "@babel/preset-typescript": "^7.6.0",
     "@babel/register": "^7.6.2",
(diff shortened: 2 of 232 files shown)


steveluscher:
Hey @ianstormtaylor, do you have a moment to smash this in? I'd like to release and announce the Solana React Native SDK but this is the last thing preventing me from giving people a good ‘Quick Start’ experience without extra configuration.

steveluscher:
Pretty please? I _really_ want to be able to remove this patch as a line in my Getting Started README :)

amaranth:
Why does this also change the test setup and all the tests?

steveluscher:
I made this PR so long ago that the only thing I remember about the changes to the test system is what I wrote in the PR summary itself.

ianstormtaylor:
Hey @steveluscher, very sorry for the delay here. Thank you for putting this together, and makes sense to test directly from source.

ianstormtaylor:
Sorry I should have looked more closely, but I've reverted this change. This is an upstream issue (`.cjs` and `.mjs` files should be recognized by Metro) and it looks like it has been fixed upstream.

steveluscher:
Haha. True. Since I made this change, Metro got its shit together. Thanks anyway!

PR #1105, opened 2022-09-01
fix coercion example in README

sisp:
I've fixed a minor error in the coercion example in the README.


ianstormtaylor:
Thanks @sisp good catch, I always confuse myself with that one! :smile:

Issue #981, opened 2021-12-01
optional() result on returning null after object creation


aldipower:
Given this struct:
```JavaScript
const User = object({
  userId: string(),
  description: optional(string()),
});
```

creating an object without a description

```JavaScript
const user = create({ userId }, User);
```

results in an object with a description of `null`
```JavaScript
{
  userId: "sadfsdfxcbvxd",
  description: null,
}
```

I would expect `undefined` here for the `description`.
`null` adds actually a value and changes the structure of the object by adding a key.

Is this intentional?
This feels like a bug to me.

yss14:
@aldipower For me your use case returns `undefined`, see [example](https://stackblitz.com/edit/node-wyas5t).

However, I would not expect the property to be set `undefined` when the input object doesn't own the property at all.

PR #1106, opened 2022-09-14
Stop coercing empty optional fields to undefined values

cwouam:
While using create to coerce some data, optional fields are defined to undefined values.
This causes problems when storing the result directly in MongoDB (generates null values).
The PR removes these undefined fields from the result of the create call.

Fix for #981


ianstormtaylor:
Great call, thanks @cwouam!

Issue #1111, opened 2022-09-20
Error on using a simple example


chuacw:
Hi, I'm using superstruct in a simple project, and got the following error when trying to run it with nodejs.

```
Error [ERR_REQUIRE_ESM]: require() of ES Module C:\Users\cheewee-chua-laptop\Development\teststruct\node_modules\superstruct\lib\index.cjs.js from C:\Users\cheewee-chua-laptop\Development\teststruct\teststruct.ts not supported.
index.cjs.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.cjs.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in C:\Users\cheewee-chua-laptop\Development\teststruct\node_modules\superstruct\package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

package.json
```
{
  "name": "teststruct",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "superstruct": "^0.16.1"
  },

  "devDependencies": {
    "ts-node": "^10.9.1",
    "typescript": "^4.8.3"
  }
}
```

tsconfig.json
```
{
  "compilerOptions": {
    "target": "es2016",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
    "module": "commonjs",                                /* Specify what module code is generated. */
    "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */

    "esModuleInterop": true,                             /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables 'allowSyntheticDefaultImports' for type compatibility. */
    // "preserveSymlinks": true,                         /* Disable resolving symlinks to their realpath. This correlates to the same flag in node. */
    "forceConsistentCasingInFileNames": true,            /* Ensure that casing is correct in imports. */

    /* Type Checking */
    "strict": true,                                      /* Enable all strict type-checking options. */
    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
  }
}

```

teststruct.ts
```
import { create, object, number, string, defaulted } from 'superstruct';

let i = 0

const User = object({
[...]

PR #1112, opened 2022-09-21
Support Node 14 (again)

mcmire:
It appears that between 0.16.0 and 0.16.1, the minimum version of Node required to use this package changed, from 14.x to 16.x. This was not explicit but seems to have been caused by a couple of factors.

But first, what changed. If you look at `src/error.ts` in 0.16.0 you will see [this line][1]:

```
return (cached ??= [failure, ...failures()])
```

In the [published version of this file in 0.16.0][2] this gets transpiled to:

```
return (_cached = cached) != null ? _cached : cached = [failure, ...failures()];
```

In 0.16.1, the [original line is unchanged][3], but in the [published version][4] it is transpiled to:

```
return cached ??= [failure, ...failures()];
```

The `??=` syntax is not supported by Node 14, hence, developers are forced to upgrade to at least Node 16 if they want to use v0.16.1 or greater.

After looking at the diff between these two versions and running some experiments, I believe that there are two reasons why this line shows up differently in these two published versions.

1. Different Node versions were used to build and publish these versions. It appears that Node 14 was used for the former whereas Node 16 was used for the latter. This assertion is supported by the fact that in the [Rollup configuration][5], `@babel/preset-env` is configured with `node: true`, which instructs Babel to [use the current version of Node as a target][6]. So if the current Node version changes, so does the Babel config.
2. Between 0.16.0 and 0.16.1, [`browserslist` was updated from 4.20.3, to 4.21.4][7] (you will probably need to expand `yarn.lock`; if so, Cmd-F for "browserslist"). In `browserslist` 4.21.0, [IE 11 was removed from the default set of browsers][8] (which is being used in this case, since no explicit list is provided). According to caniuse, [IE 11 does not support the `??=` syntax][9], so its removal means that Babel doesn't need to transpile this syntax any longer.

To address this problem, this PR:

* changes the Rollup configuration mentioned above to use `node: "14.0"` instead of `node: true`, so that Node 14 is always used to compute the transpilation rules regardless of the version of Node used locally to build and publish the package
* updates the CI workflow to ensure that Node 14 is being tested (along with 16 and 18)
* adds Node >= 14 to the `engines` field to communicate that it is supported

---

One thing you may wonder is why this change is needed at all. Node 16 is the current LTS, so shouldn't that be enough? True, but Node 14 hasn't hit end-of-life yet, and many people are still using it, including my company. We think this package is really great, but it would be even better if we didn't have to have a workaround for our libraries that we still want to keep on Node 14.

Thanks for considering :)

[1]: https://github.com/ianstormtaylor/superstruct/blob/v0.16.0/src/error.ts#L44
[2]: https://unpkg.com/superstruct@0.16.0/lib/index.cjs
[3]: https://github.com/ianstormtaylor/superstruct/blob/v0.16.1/src/error.ts#L44
[4]: https://unpkg.com/superstruct@0.16.1/lib/index.cjs.js
[5]: https://github.com/ianstormtaylor/superstruct/blob/v0.16.4/config/rollup.js#L26
[6]: https://babel.dev/docs/en/options#targetsnode
[7]: https://github.com/ianstormtaylor/superstruct/compare/v0.16.0...v0.16.1?diff=unified#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL99
[8]: https://github.com/browserslist/browserslist/blob/main/CHANGELOG.md#421
[9]: https://caniuse.com/?search=%3F%3F%3D

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,14 @@ on: [push, pull_request]
 jobs:
   ci:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v2
         with:
-          node-version: lts/*
+          node-version: ${{ matrix.node-version }}
       - run: yarn && yarn test && yarn lint
         env:
           CI: true
--- a/config/rollup.js
+++ b/config/rollup.js
@@ -23,7 +23,7 @@ export default {
         [
           '@babel/preset-env',
           {
-            targets: { browsers: 'defaults, not ie 11', node: true },
+            targets: { browsers: 'defaults, not ie 11', node: '14.0' },
             modules: false,
             useBuiltIns: false,
             loose: true,
--- a/package.json
+++ b/package.json
@@ -107,5 +107,8 @@
     "validation",
     "validator"
   ],
-  "dependencies": {}
+  "dependencies": {},
+  "engines": {
+    "node": ">=14.0.0"
+  }
 }


mcmire:
This is a carryover from the comment I left in this issue: https://github.com/ianstormtaylor/superstruct/issues/1111#issuecomment-1253903140. I know originally I was going to submit a PR that set `engines` to `>= 16.0.0`, but after discussing with my team we felt like it would be more beneficial to not only us, but also everyone, if `superstruct` could support Node 14 in addition to the versions it supports now. Let me know what you think :)

ianstormtaylor:
@mcmire this is an absolutely beautiful description and pull request. Thank you! I completely agree with everything you said. And I'm sorry for the regression.

mcmire:
Super, no problem. Thanks for the fast merge!

Issue #1095, opened 2022-08-10
Refinement function return types are not documented


arturmuller:
Hi — I noticed that the docs don't mention the ability to return strings, Failures, etc... from refinement functions to customize error messages. Is this intentional (ie. is this just for internal use?) or just an omission?

I am happy to add it if there's interest.

ianstormtaylor:
@arturmuller would love a PR! Good call.

PR #1113, opened 2022-09-22
Document custom error messages for refinements

arturmuller:
As per https://github.com/ianstormtaylor/superstruct/issues/1095.

Note that I also added a short "robot tip" about the typing of refiner function's `value`, as I felt it might be reassuring to know that `value.startDate < value.endDate` is guaranteed to be safe.


ianstormtaylor:
Thank you @arturmuller, this is awesome!

PR #1114, opened 2022-09-27
Allow optional and nullable enums

TomPenguin:
I'd like to add `nullable` in [this](https://github.com/ianstormtaylor/superstruct/pull/694) case.

I couldn't find any problems with following:

```bash
yarn test
```

and

```ts
const undefinedTestTypeSchema: Describe<undefined> = literal(undefined)

const nullTestTypeSchema: Describe<null> = literal(null)

const optionalTestTypeSchema: Describe<string | undefined> = optional(string())

const partialEnumTypeSchema: Describe<{ foo?: 'bar' | 'baz' }> = object({
  foo: optional(enums(['bar', 'baz'] as const)),
})

const partialAndNullableEnumTypeSchema: Describe<{
  foo?: 'bar' | 'baz' | null
}> = object({
  foo: optional(nullable(enums(['bar', 'baz'] as const))),
})

const nullableEnumTypeSchema: Describe<{
  foo: 'bar' | 'baz' | null
}> = object({
  foo: nullable(enums(['bar', 'baz'] as const)),
})

const optionalUndefinedSchema: Describe<{ foo?: undefined }> = object({
  foo: literal(undefined),
})

const optionalNullSchema: Describe<{ foo?: null }> = object({
  foo: optional(literal(null)),
})
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -213,7 +213,7 @@ export type Assign<T, U> = Simplify<U & Omit<T, keyof U>>
  * A schema for enum structs.
  */
 
-export type EnumSchema<T extends string | number | undefined> = {
+export type EnumSchema<T extends string | number | undefined | null> = {
   [K in NonNullable<T>]: K
 }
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @TomPenguin! Can you add some typings tests to account for the new allowed behavior?

PR #1126, opened 2022-10-22
Allow coercions to undefined

### LINKED ISSUE #1125: No longer possible to coerce values to `undefined`
If I return `undefined` from a coercion, the coercion is ignored completely, and the original value is kept (even if the original value does not satisfy the struct we're coercing the value into).

The most common use case in my project is:

```ts
const coercableOptional = <T>(struct: Struct<T>) =>
  coerce(optional(struct), literal(null), () => undefined);
```

This no longer works: `null` values are retained, even if `struct` should not allow null values.

I think this was caused by #1106

Is there an alternative syntax for coercing to undefined, or is this a bug that needs fixing?

benj-dobs:
Closes #1125 

Ensures explicit undefined values in coercions are respected. (See linked issue for details)

This is my first time contributing, so let me know if you'd like anything done differently.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -175,7 +175,7 @@ export function* run<T, S>(
         } else if (value instanceof Set) {
           value.add(v)
         } else if (isObject(value)) {
-          if (v !== undefined) value[k] = v
+          if (v !== undefined || k in value) value[k] = v
         }
       }
     }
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks @benj-dobs, good catch here!

PR #1135, opened 2022-11-01
Fix Symbol serialisation

### LINKED ISSUE #1134: Superstruct cannot create error message if a Symbol fails validation
When a `Symbol` fails validation, Superstruct is unable to create an error message for it. This results in Superstruct throwing an error message, even when using `is` for validation.

This can be reproduced using:

```ts
is(Symbol('foo'), string());
```

The expected result is `false`, but instead this throws the following error:

```
TypeError: Cannot convert a Symbol value to a string
    at value (./node_modules/superstruct/src/utils.ts:38:65)
    at print (./node_modules/superstruct/src/structs/types.ts:423:43)
    at Struct.validator (./node_modules/superstruct/src/struct.ts:46:24)
```

Looks like Superstruct is trying to create a stringified version of it in this function, which is not possible for `Symbol`s:

https://github.com/ianstormtaylor/superstruct/blob/fe7e79779573c6619265013919f913e7c6c2138c/src/utils.ts#L37-L39

Mrtenz:
Fixes #1134.

Doing \`${Symbol('foo')}\` results in "TypeError: Cannot convert a Symbol value to a string". Using `Symbol('foo').toString()` works however, and results in the string `"Symbol(foo)"`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -35,6 +35,10 @@ export function isPlainObject(x: unknown): x is { [key: string]: any } {
  */
 
 export function print(value: any): string {
+  if (typeof value === 'symbol') {
+    return value.toString()
+  }
+
   return typeof value === 'string' ? JSON.stringify(value) : `${value}`
 }
 


ianstormtaylor:
Good catch, thank you @Mrtenz!

Issue #1096, opened 2022-08-17
Runtime error when using create() on superstruct.record with a frozen object


gwisp2:
### Code
[codesandbox](https://codesandbox.io/s/mystifying-chandrasekhar-wj624d?file=/src/index.ts)
```typescript
import { record, string } from "superstruct";

const exampleStruct = record(string(), string());
console.log(exampleStruct.create(Object.freeze({ a: "b" })));
```

### Expected result
```javascript
{ "a": "b" }
```

### Actual result
TypeError
Cannot assign to read only property 'a' of object '#<Object>'
[at utils.ts#L178](https://github.com/ianstormtaylor/superstruct/blob/d1508a9e0f54723e463eeed8ab856dbd9560cd72/src/utils.ts#L178)

### Possible cause
`object()` creates a Struct whose `coercer` method creates a copy of the input.
`record()` creates a Struct with a default `coerce` method that returns the input without changes.

PR #1138, opened 2022-11-07
Deeply coerce within unions

### LINKED ISSUE #971: Coercions nested inside object doesn't work

### LINKED ISSUE #1136: Union doesn't work as expected when it contains objects with coercions
I would have expected the below to pass and coerce, but got a struct error instead.

```ts
const struct = union([
  object({
    type: literal("A"),
    value: defaulted(enums(["a", "b", "c"]), "a"),
  }),
]);

const value = create({ type: "A" }, struct); // Throws StructError
```

It seems to be a specific issue with objects within unions, as the below works fine:

```ts
const struct = union([defaulted(enums(["a", "b", "c"]), "a")]);
const value = create(undefined, struct); // Works!
```

@ianstormtaylor lmk if from your perspective this should be fixed — if so, I am happy to take a crack at it and open a PR.

arturmuller:
Hey @ianstormtaylor, as per #1136

**👉 TL;DR:** This PR fixes the linked issue (coercion of object properties within unions), but perhaps some next steps should be taken to fix similar issues with other structs. More below -- would love to hear your thoughts!

## The Bug

The [coercer for `union`](https://github.com/ianstormtaylor/superstruct/blob/6717743ee286b13dde6de072019194afa75bfb29/src/structs/types.ts#L499-L506) seems to be relying on a mutation of input values which happens during the execution of `s.validate()`, specifically [this line](https://github.com/ianstormtaylor/superstruct/blob/6717743ee286b13dde6de072019194afa75bfb29/src/utils.ts#L182) inside the `run` util.

The [existing tests](https://github.com/ianstormtaylor/superstruct/blob/6717743ee286b13dde6de072019194afa75bfb29/test/validation/union/coercion-nested.ts) for nested coercions pass because they use `type`. Unlike other structs (`object`, `map`, `set`, etc...), `type` does not copy its value inside its coercer -- it uses the [default struct coercer](https://github.com/ianstormtaylor/superstruct/blob/6717743ee286b13dde6de072019194afa75bfb29/src/struct.ts#L35), which just returns its input.

So, the union coercer code below has pretty different behaviour depending on whether the coercer for the given struct copies it's value or not.

```ts
const firstMatch =
  Structs.find((s) => {
    const [e] = s.validate(value, { coerce: true }) // [1] Mutative deep coercion incl. entries
    return !e
  }) || unknown()

return firstMatch.coercer(value, ctx) // [2] Shallow coercion
```

The above works for `type` because it mutates `value` during the deep coercion step at **[1]** and even thought step **[2]** doesn't do deep coercion it still arrives at the correct result because of the previous mutation.

For an `object` struct, step **[1]** correctly finds the first matching struct, but step **[2]** does not take `entries` into account and, since value is copied in this case, a value with uncoerced entries is returned. This uncoerced value is what is later caught by the validation step.

## PR Changes

1. The `union` coercer now returns fully coerced values
2. Adds coercer to the `type` struct identical to the `object` struct (shallowly copies input).
3. Updates tests

One thing to note: the coercer previously passed context to `struct.coerce(value, ctx)`. This is no longer present — it doesn't seem to do anything and other coercers don't use it, but just wanted to flag it in case there in fact was a reason that I am not seeing.

## Next Up

If accidental mutation is not welcome (FWIW, my personal expectation always has been that values passed to superstruct won't be mutated -- that said, the documentation doesn't make any such promises I think), it would probably make sense to double check whether other structs also potentially cause mutation. My hunch is that #1096 is caused exactly by this issue.

---

Resolves #1136, resolves #971, closes #1038


arturmuller:
For additional context:

PR #1038 is addressing the same bug. I ran the tests presented in his PR on top of the changes introduced here and they all pass.

I also (now -- after the fact, ooops 😅) read throught #971 and the related convo by @PoyangLiu and @yss14. FWIW based on my experiments, a change to `entries` does not resolve the issue. That said, I am absolutely ready to be proven wrong! 😄

DavidHudlow:
Doesn't seem to fix cases where the coercer is buried below the union.

```

const ValidatedFarm = object({
  farmType: enums(['dairy', 'egg']),
  grass: coerce(boolean(), string(), (value) => {
    return value === 'true'
  }),
})

const CowFarm = object({
  ...ValidatedFarm.schema,
  farmType: literal('dairy'),
  cow: boolean(),
})

const EggFarm = object({
  ...ValidatedFarm.schema,
  farmType: literal('egg'),
  chicken: boolean(),
})

type Farm =
  | {
      farmType: 'dairy'
      grass: boolean
      cow: boolean
    }
  | {
      farmType: 'egg'
      grass: boolean
      chicken: boolean
    }

const ValidatedAnyFarm: Struct<Farm> = union([CowFarm, EggFarm])
const ValidatedAnyFarms: Struct<Farm[]> = array(ValidatedAnyFarm)

describe('Unions', () => {
  it('allows chicken type', () => {
    ValidatedAnyFarm.assert({
      chicken: true,
      farmType: 'egg',
      grass: false,
    })
  })
  it('allows cow type', () => {
    ValidatedAnyFarm.assert({ cow: true, farmType: 'dairy', grass: true })
  })
  it('allows chicken type with coercion required', () => {
    ValidatedAnyFarm.assert({
      chicken: true,
      farmType: 'egg',
      grass: 'false',
    })
  })
  it('allows cow type with coercion required', () => {
    ValidatedAnyFarm.assert({
      cow: true,
      farmType: 'dairy',
      grass: 'true',
    })
  })
  it('allows cow and chicken types in array', () => {
    ValidatedAnyFarms.assert([
      { cow: false, farmType: 'dairy', grass: true },
      { chicken: true, farmType: 'egg', grass: false },
    ])
  })
})
```

Test fails ^

arturmuller:
> EDIT: Had an error in my tests, but after fixing it, this does pass the tests I've been running!

@DavidHudlow are these tests accessible somewhere? I'd love to see them for reference!

DavidHudlow:
Sure, here's the fixed version:

```
import { deepStrictEqual } from 'assert'
import {
  string,
  object,
  union,
  coerce,
  enums,
  literal,
  boolean,
  array,
  number,
} from '../../src'

const ValidatedFarm = object({
  farmType: enums(['dairy', 'egg']),
  grass: coerce(boolean(), string(), (value: any) => {
    return value === 'true'
  }),
})

const CowFarm = object({
  ...ValidatedFarm.schema,
  farmType: literal('dairy'),
  cows: number(),
})

const EggFarm = object({
  ...ValidatedFarm.schema,
  farmType: literal('egg'),
  chickens: number(),
})

const ValidatedAnyFarm = union([CowFarm, EggFarm])

const cowFarm = { cows: 20, farmType: 'dairy', grass: true }
const chickenFarm = { chickens: 12, farmType: 'egg', grass: false }
const cowFarmNeedsCoerce = { cows: 62, farmType: 'dairy', grass: 'true' }
const chickenFarmNeedsCoerce = { chickens: 6, farmType: 'egg', grass: 'false' }

describe('Union with deep coercion', () => {
  it('allows chicken type', () => {
    deepStrictEqual(ValidatedAnyFarm.validate(chickenFarm)[0], undefined)
    deepStrictEqual(ValidatedAnyFarm.validate(chickenFarm)[1], chickenFarm)
  })
  it('allows cow type', () => {
    deepStrictEqual(ValidatedAnyFarm.validate(cowFarm)[0], undefined)
    deepStrictEqual(ValidatedAnyFarm.validate(cowFarm)[1], cowFarm)
  })
  it('allows chicken type with coercion required', () => {
    deepStrictEqual(
      ValidatedAnyFarm.validate(chickenFarmNeedsCoerce, { coerce: true })[0],
      undefined
    )
    deepStrictEqual(
      ValidatedAnyFarm.validate(chickenFarmNeedsCoerce, { coerce: true })[1],
      {
        ...chickenFarmNeedsCoerce,
        grass: false,
      }
    )
  })
  it('allows cow type with coercion required', () => {
    deepStrictEqual(
      ValidatedAnyFarm.validate(cowFarmNeedsCoerce, { coerce: true })[0],
      undefined
    )
    deepStrictEqual(
      ValidatedAnyFarm.validate(cowFarmNeedsCoerce, { coerce: true })[1],
      {
        ...cowFarmNeedsCoerce,
        grass: true,
      }
    )
  })
  it('allows cow and chicken types in array', () => {
    const [errors, value] = array(ValidatedAnyFarm).validate(
      [cowFarm, chickenFarm, cowFarmNeedsCoerce, chickenFarmNeedsCoerce],
      { coerce: true }
    )
    deepStrictEqual(errors, undefined)
    deepStrictEqual(value[2], { ...cowFarmNeedsCoerce, grass: true })
  })
})
```

ianstormtaylor:
Thank you @arturmuller, this is great!

PR #1149, opened 2022-11-18
Retain struct's type in partial and pick helpers

### LINKED ISSUE #829: omit, pick and partial should return schema's type

arturmuller:
Resolves #829

### DIFF
--- a/docs/reference/utilities.md
+++ b/docs/reference/utilities.md
@@ -92,7 +92,7 @@ partial(
 { name: 'Jane' }
 ```
 
-`partial` allows you to create a new struct based on an existing object struct, but with all of its properties being optional.
+`partial` allows you to create a new struct based on an existing `object` or `type` struct, but with all of its properties being optional.
 
 ### `pick`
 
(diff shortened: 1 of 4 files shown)


dlindenkreuz:
@ianstormtaylor Are there any blockers? This would be a really useful fix.

kevcao-certik:
@ianstormtaylor Same here, this would be super useful if there are no issues with the PR. Thanks!

arturmuller:
@dlindenkreuz @kevcao-certik this will be landing in the next release!

I will be helping Ian with the maintenance of Superstruct from now on, so as soon as I get publish rights on npm you will be able to (finally) use this :)

PR #1165, opened 2023-01-23
make more of the examples runnable

rauno56:
Many of the examples in the docs have some imports but not all. That was confusing for me, especially because `assert` exists in std lib of Node.js.

This PR doesn't thrive do make all examples copy-pasteable, only the ones that already had import declarations by making the list of imports complete.


arturmuller:
Thanks @rauno56 — this is great! Really appreciate it ✌️ 

I am waiting for Ian to give me access to updating the docs site — this will be live as soon as I get it.

PR #1181, opened 2023-05-10
feat: add `InferUncoerced` type utility

ziad-saab:
Works exactly like `Infer`, except where part of the schema is coerced, returns the un-coerced type instead. Tests passing.


amok:
Any chance it gets merged?

Issue #803, opened 2021-05-21
Mask with unions


lvanhala:
Is there a way to make `mask` work with `union`? I'd like to define only one parser that I could use to:
- Validate object and make sure it does not contain any extra fields
- Parse object ignoring extra fields

I know I could use `type()` instead of `object()` but that would essentially require defining two parsers for the whole huge object tree.

Here's the minimal use case:

```
import { object, string, mask, literal, union } from 'superstruct'

const ItemA = object({
  type: literal('a'),
  valueA: string(),
})

const ItemB = object({
  type: literal('b'),
  valueB: string(),
})

// Works
mask(
  {
    type: 'a',
    valueA: 'test',
    extraField: true,
  },
  ItemA
)

// Fails
mask(
  {
    type: 'a',
    valueA: 'test',
    extraField: true,
  },
  union([ItemA, ItemB])
)

```

dimikot:
@ianstormtaylor I believe the behavior mentioned above is actually a bugreport. Just prepared a failing test case for it independently, and then found the current issue reported:

```
  it.only('masking succeeds validation for objects with extra props in union', () => {
    const S = union([object({ a: string() }), object({ b: string() })])
    const value = { a: '1', extraProp: 42 }
    deepStrictEqual(mask(value, S), { a: '1' })
  })
```

arturmuller:
Resolved in 2.0

PR #1196, opened 2023-11-29
Fix mask() working incorrectly with union() when an alternative object contains extra unknown props

dimikot:
Fixes https://github.com/ianstormtaylor/superstruct/issues/803

Basically, it's a more proper rework of my previous PR https://github.com/ianstormtaylor/superstruct/pull/629 merged several year ago. The fact is that "mask" approach applies to `object` structs only, so it's more proper even conceptually to put the masking logic into object() constructor rather than having it in run(). By doing so, I also fix a bug where union() did not work properly with mask(): before, instead of removing the unknown props when a union element matches, it mistakenly threw a "Expected the value to satisfy a union" error.

## Test plan

```
npm run test
```

Before, the added test threw:

CleanShot 2023-11-28 at 18 39 03@2x

After, all the tests (including the new ones) pass.

### DIFF
--- a/src/struct.ts
+++ b/src/struct.ts
@@ -86,7 +86,8 @@ export class Struct<T = unknown, S = unknown> {
 
   /**
    * Mask a value, coercing and validating it, but returning only the subset of
-   * properties defined by the struct's schema.
+   * properties defined by the struct's schema. Masking applies recursively to
+   * props of `object` structs only.
    */
 
   mask(value: unknown, message?: string): T {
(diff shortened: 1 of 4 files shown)


dimikot:
@ianstormtaylor could you please take a look? This is a pretty straighforward change.

dimikot:
@ianstormtaylor 🙏🙏

arturmuller:
Hey @dimikot — I have recently spoken to Ian and I will be helping out with the maintenance of Superstruct from now on.

Getting `union` to work with `mask` has actually been on the top of my priority list (it's so useful and a real bummer that it currently doesn't work!) so I am psyched that you've already taken a crack at it! ✌️ 

I will look into your PR closely next week and get back to you.

arturmuller:
Ok, looked through your PR: One thing I am not sure about is whether it a problem that the output of `mask` with opaque `object` has changed if an array is passed as value.

```ts
// Before
mask([0, 1, 2], object()) // Returns { '0': 0, '1': 1, '2': 2 }

// Your PR
mask([0, 1, 2], object()) // Returns [ 0, 1, 2 ]
```

Admittedly, it doesn't feel like a very common use-case, and I actually think the new behaviour is more correct, but I would hate to have some patch-version-bump breakages for people in my first couple of releases.

dimikot:
I think I broke it in the old PR https://github.com/ianstormtaylor/superstruct/pull/629 actually (this is likely where `{ '0': 0, '1': 1, '2': 2 }` behavior was introduced first), so in the current PR, it's returning back to the roots. But I hear you, I'll try to figure out, how to keep the current behavior, even though it's a little unexpected.

arturmuller:
Hi @dimikot — please lmk if you would like to finish this PR. 🙏

arturmuller:
@dimikot — thank you for your work on this. I've merged it into `main` in a separate PR as I wasn't able to make changes in your branch, but your commit (and most importantly the fix itself) have made it into 2.0! Thank you so much 🙌

PR #1211, opened 2024-01-28
Module resolution, switched testing to vitest

ciscoheat:
This modernizes the package in a few ways:

- Adding ".js" extensions to import statements, for modern bundler compatibility.
- Switched to [vitest](https://vitest.dev) testing framework, to get rid of babel.


MentalGear:
@ianstormtaylor Could you please review this PR - would allow for Sveltekit's most popular Form Validation Lib to use superstruct !

arturmuller:
Hi @ciscoheat — thank you so much for this PR — it is hugely appreciated! 🎉 

Getting Superstruct to be compatible with NodeNext is my priority right now (I have recently taken over maintenance of Superstruct from Ian).

Given that Superstruct has quite a few downloads on `npm`, I would really like to make sure that versioning properly follows semver and we don't introduce some nasty surprises to unsuspecing users.

There are multiple interlocking issues going on here (node resolution, typescript config, mocha/testing) and I think we have to tackle them one at a time.

I made a new PR #1224 where I would *love* your input of the actual NodeNext compat. Importantly, if you want to try out the file-extensions solution to use superstruct with NodeNext *now*, you can install `superstruct@1.0.5-0` prerelease version.

ciscoheat:
Hi Artur, thank you, I know I should have separated the two things, but they became intermingled during experimentation. I'm mostly concerned with NodeNext compat for [Superforms](https://superforms.rocks/), so I'll try the prerelease version and see how it goes! About #1226 I can separate the PR to only focus on vitest, but I'm afraid I cannot do much more with the time I have right now. Should I base it on the prerelease #1224?

arturmuller:
Thank you for getting this off the ground @ciscoheat — I am closing this PR since we solved it in another PR, but your work on this is still super appreciated! 💪

PR #1224, opened 2024-03-11
NodeNext Compat

arturmuller:
> 👉 This PR can be tested via a pre-release version of Superstruct: `npm i superstruct@1.0.5-0`.

The goal of this PR is to fix Superstruct's compatibility with TypeScript's [NodeNext](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext) module / module resolution settings. The simplest fix I could think of (and which was also proposed in #1211 by @ciscoheat) is to add file extensions to imports. This is currently in the PR, however, this presents several issues:

1. **It breaks tests**. Test are currently written in Mocha, which is really starting to show its age — it is quite hard to figure out how to make things play nice with the rest of the tooling. @ciscoheat proposed a switch to Vitest which I am 100% in favour of, but I think we need to make the switch a bit more thoughtfully: a) Make sure the tests are actually written in a more standard Vitest way and that all test are passing.
2. We have to actually use the file extensions, which — in my experience — is rather confusing to new contributors. The extension has to be `.js` not `.ts` (since TS doesn't rewrite paths), and we lose directory import. Bottom line, this feels like a non-ideal way to do this and other libs manage in a more elegant way, so we investigate how this can be done.
3. There are many environments where Superstruct can run, and any changes to module resolution should be **either thoroughly tested** (at least manually), **or marked as a breaking change** to make sure we don't break suddenly people's builds. I would really hate to drop the whole TS/Node/CommonJS/ESM mess on unsuspecting users.


ciscoheat:
Hello again @arturmuller, any progress on this one? It would be very nice to support another library for Superforms.

Jakeii:
I suggest adding [`tsc-alias`](https://github.com/justkey007/tsc-alias) to the build process, you can leave the extensions out of the ts files, and just add the `tsc-alias` command after the build command.

with
```json
  "tsc-alias": {
    "resolveFullPaths": true
  }
```

in the tsconfig.

it'll add extensions to all the paths in the type definitions.

Jakeii:
> I suggest adding [`tsc-alias`](https://github.com/justkey007/tsc-alias) to the build process, you can leave the extensions out of the ts files, and just add the `tsc-alias` command after the build command.
> 
> with
> 
> ```json
>   "tsc-alias": {
>     "resolveFullPaths": true
>   }
> ```
> 
> in the tsconfig.
> 
> it'll add extensions to all the paths in the type definitions.
> 
> PR: #1232

Hey @arturmuller what do you think?

yeoffrey:
This could be rebased onto main now that the tests are running on something newer!

PR #1232, opened 2024-05-03
add tsc-alias for nodenext compatibility

Jakeii:
Use `tsc-alias` to add file extensions to type definitions for compatibility with `NodeNext`

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "np": "^10.0.0",
     "prettier": "^3.2.5",
     "rollup": "^4.12.1",
+    "tsc-alias": "^1.8.8",
     "typescript": "^4.8.3"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


arturmuller:
Hi @Jakeii — thank you so much for your PR!

I tested this locally, and unfortunately it doesn't seem to actually fix our issue.

Our core problem is with types declarations, not actual source code (since that's bundled and therefore there are no imports) and I am not sure if tsc-alias actually hooks into that.

Please lmk if you're still interested in trying to get this up and running.

@yeoffrey managed to unblock us by moving over to Vitest so the road to NodeNext compat has been cleared and this is not blocking for the release (any more).

Either way — thank you so much for the contrib! 🙌

arturmuller:
@Jakeii I am going to close this for now, but if you feel like taking another crack at removing the .js extensions from our import paths in typedefs a PR is very welcome!

Jakeii:
Hi @arturmuller, the `tsc-alias` plugin removes the need to add `js` extensions in `ts` source files, it adds them for you in both the generated `js` and `d.ts` files.

I did test locally and in worked for me?

PR #1237, opened 2024-05-24
Document pass-by-reference for `defaulted` values

### LINKED ISSUE #1236: [Docs] Document pass-by-reference pitfall for `defaulted` values
I just spent about half an hour trying to figure out why data was showing up in places it shouldn't. It turned out that my `defaulted` values were being passed by reference, meaning that any modifications to one instance would affect all current and future instances `create`d from the same struct.

```ts
// BAD: array is passed by reference
const MyType = defaulted(array(string()), []);

const instanceA = create(undefined, MyType);

instanceA.push("hello");

const instanceB = create(undefined, MyType);
console.log(instanceB);
// [ "hello" ]
```

While this makes sense in hindsight, it was pretty painful to debug (especially since my case had a significantly more complex data structure).

As such, I think this project would be improved if this pitfall was documented. I've created a PR that adds this documentation.

Additionally, I noticed that the type signatures for defaulted allow `any` as the default value. Is this intentional? I imagine that stricter types would allow for significantly improved safety, helping users to avoid this issue (eg only accepting a plain value if the value is a type that cannot be modified).

MiguelGuthridge:
Fixes #1236

Not 100% sure if this is the best place to place the documentation. Also not sure if this level of thoroughness is too much.

Let me know if you'd like me to move the documentation elsewhere, or make it more brief.

### DIFF
--- a/docs/guides/03-coercing-data.md
+++ b/docs/guides/03-coercing-data.md
@@ -46,6 +46,35 @@ const User = defaulted(
 )
 ```
 
+### Pass-by-reference
+
+While providing a plain value for a `defaulted` property does work, the value will be passed by reference. This means that any modifications to the value will affect other instances that receive the reference.
+
+```ts
+const NamesBuggy = defaulted(array(string()), []);
+// BAD: [] will be passed by reference        ^
+
+const names1 = create(undefined, NamesBuggy);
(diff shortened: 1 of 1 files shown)


MiguelGuthridge:
Issue is actually more complex, will close this

Issue #1226, opened 2024-03-11
Move tests to Vitest


arturmuller:
Superstruct tests are currently written in Mocha, which is starting to show its age. It would be great if we could move to something like Vitest. This has two goals:

1. Unblock #1224 (Mocha is currently failing because it is not compatible with the new config)
2. Make the lib more comfortable to contribute/work with for new contributors — I doubt there are many people who would choose Mocha over Vitest in 2024.

@ciscoheat — I saw that you already got started on this in #1211 , but when I try to run it locally it didn't actually run most of the tests. If you were willing to look into this again, with the sole focus being moving to Vitest, it would be massively appreciated. 🙌 

So that goal is:

1. Making sure that **all** tests run
2. (stretch goal) Refactoring the tests so that they seem familiar to a person who has only ever worked with Vitest. (The current implementation is not how I would expect anyone to write Vitest tests).

PR #1242, opened 2024-06-02
Migrate testing suite to Vitest

yeoffrey:
Closes: #1226 

# Description
This PR contains the changes nessessary to migrate the current testing suite, which is Mocha to Vitest. This is similar to [ciscoheat's PR](https://github.com/ianstormtaylor/superstruct/pull/1211), except that it only contains Vitest changes.

Note: Vitest automatically finds tests by looking for filenames that end with `.test.ts`, so I've done this here. This should also help unblock #1224.

# Next steps
I believe that the current way which the test suite gathers files from the file system is not ideal. I think theres a better way to organize these tests, but that should be in another PR.


test/index.ts
```diff
@@ -1,129 +1,5 @@
...
-} from '../src'
-
-describe('superstruct', () => {
-  describe('api', () => {
-    require('./api/assert')
-    require('./api/create')
-    require('./api/is')
-    require('./api/mask')
-    require('./api/validate')
-  })
```
yeoffrey, line 21: Since these are automatically run by changing the filename to end with `.test.ts`, I've removed them here.

arturmuller:
Hey @yeoffrey -- this is lovely! ❤️ 

I agree that to make things feel more Vitest-y it should probably be organised differently. PRs on that front extremely welcome!

Merging...

Issue #1244, opened 2024-06-09
Simplify testing suite with Vitest


yeoffrey:
# TL;DR
The current testing suite is overly complicated. Since we've moved over to Vitest (see: #1242), we should now that package to its full potential, and simplify the testing suite all together for this project.

# More Detail
Currently the testing suite basically has a "runner" which is in `index.test.ts` which handles all the tests in the testing suite. It imports configurations from `/validation` which will then be imported with the runner (not by a typescript import, but by a module import 🤢 ) and it will execute tests.

The problem here is that is super complex. It also feels like a blackbox a bit because its not super readable and its not very strict. Now that we are using Vitest, I think theres a great opportunity to make this easier to understand and also more reliable.

# Goals
- Report code coverage
- Allow users to use the `watch` feature while testing
- Make the testing suite easy to read and add to.

PR #1245, opened 2024-06-10
Use Vitest and expect methods for testing suite

yeoffrey:
In this PR, the test suite has been modified:
- `deepStrictEqual` and `deepEqual` have been changed to `.toStrictEqual` for objects and arrays and `.toBe` for literals.
- Moved the deprecated tests to their own file.
- I've opted to keep the usage of `throws` and `doesNotThrow` because it offers a nicer api to check whether an error was thrown or not than the Jesty `expect` solution.
- [Expect documentation from Vitest](https://vitest.dev/api/expect.html) for your reference!

#1244

### DIFF
--- a/test/api/validate.test.ts
+++ b/test/api/validate.test.ts
@@ -1,5 +1,4 @@
-import { deepStrictEqual, strictEqual } from 'assert'
-import { describe, it } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import {
   validate,
   string,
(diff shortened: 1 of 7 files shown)


test/api/validate.test.ts
```diff
@@ -13,20 +12,20 @@ import {
...
   })
 
   it('invalid as helper', () => {
     const S = string()
     const [err, value] = validate(42, S)
-    strictEqual(value, undefined)
-    strictEqual(err instanceof StructError, true)
-    deepStrictEqual(Array.from((err as StructError).failures()), [
+    expect(value).toStrictEqual(undefined)
+    expect(err).toBeInstanceOf(StructError)
```
yeoffrey, line 27: Quite nice :D

arturmuller (changes requested):
This is great! Thank you @yeoffrey! 🙌 

I would change the throws/doesNotThrow calls as well. Unless we really have to, I would avoid special cases and exceptions — I doubt that new contributors would realise that we use `assert` for error handling, but Vitest for everything else.

I tested this snippet and it seems to print almost the same exact message like `assert`, so I don't think we're losing any more functionality.

```ts
it('custom error message', () => {
  expect(() => string().create(42, 'Not a string!')).toThrow(
    expect.objectContaining({
      message: 'Not a string!',
      cause: 'Expected a string, but received: 42',
    })
  )
})
```

arturmuller (approved):

PR #1247, opened 2024-06-21
Remove "is-email" from testing suite

yeoffrey:
Removing `is-email` from the testing suite. Its used as a dev dependency only when testing `refine`, and it creates some complexity because we need to type that function. Not worth it in my opinion!

Related: #1244

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.1.3",
-    "is-email": "^1.0.0",
     "is-url": "^1.2.4",
     "is-uuid": "^1.0.2",
     "lodash": "^4.17.15",
--- a/package.json
+++ b/package.json
@@ -34,8 +34,6 @@
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.1.3",
-    "is-url": "^1.2.4",
-    "is-uuid": "^1.0.2",
     "lodash": "^4.17.15",
     "np": "^10.0.0",
     "prettier": "^3.2.5",
--- a/test/validation/refine/invalid.ts
+++ b/test/validation/refine/invalid.ts
@@ -1,7 +1,6 @@
-import isEmail from 'is-email'
 import { string, refine } from '../../../src'
 
-export const Struct = refine(string(), 'email', isEmail)
+export const Struct = refine(string(), 'email', (value) => value.includes('@'))
 
 export const data = 'invalid'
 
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -34,7 +34,6 @@
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.1.3",
-    "is-email": "^1.0.0",
```
yeoffrey, line 37: `is-email` is a package that validates an email. It requires using a custom type to use, and its only used when testing `refiners`. Removing this allows us to remove `/@types`.

test/validation/refine/invalid.ts
```diff
@@ -1,7 +1,6 @@
-import isEmail from 'is-email'
 import { string, refine } from '../../../src'
 
-export const Struct = refine(string(), 'email', isEmail)
+export const Struct = refine(string(), 'email', (value) => value.includes('@'))
 
 export const data = 'invalid'
```
yeoffrey, line 5: For example, I've simply made the test for a refiner check that there is an `@` symbol in the data. Since we weren't testing whether it actually works with emails, this seems simpler and within the scope of this commit.

arturmuller:
Removing the `is-email` package probably makes sense, just to make it less confusing. I tripped over it a couple of times myself!

All the modules under `typings` are a DIY way of testing TypeScript type signatures (run via the `npm run test:types` command). We definitely don't want to remove them, but it would probably make a lot of sense to instead use [Vitest's type testing](https://vitest.dev/guide/testing-types.html) features as they are more robust in my experience.

Using Vitest would also makes it a lot clearer what is actually happening with all those files -- a major plus in my book!

yeoffrey:
> Removing the `is-email` package probably makes sense, just to make it less confusing. I tripped over it a couple of times myself!
> 
> All the modules under `typings` are a DIY way of testing TypeScript type signatures (run via the `npm run test:types` command). We definitely don't want to remove them, but it would probably make a lot of sense to instead use [Vitest's type testing](https://vitest.dev/guide/testing-types.html) features as they are more robust in my experience.
> 
> Using Vitest would also makes it a lot clearer what is actually happening with all those files -- a major plus in my book!

Cool! Will keep those then. I've refactored this PR to just contain the changes for `is-email`.

arturmuller:
@yeoffrey FYI: I noticed that the is-email was used in the examples dir, along with some other similar helper packages. I added a small README and a custom package.json to that dir so that we can remove all of these from the root.

yeoffrey:
> @yeoffrey FYI: I noticed that the is-email was used in the examples dir, along with some other similar helper packages. I added a small README and a custom package.json to that dir so that we can remove all of these from the root.

No worries at all. Thanks for doing that!

PR #1248, opened 2024-06-23
Defaulted should clone default value if it's an object

### LINKED ISSUE #1238: Defaulted values are not correctly copied for `record` type data.
When providing a default value for a `record`-type struct, the default value is not properly copied, meaning that it is at risk of subtle pass-by-reference bugs.

Consider this example:

```ts
import { create, defaulted, record, string } from 'superstruct';

// BAD: array is passed by reference
// =================================
const DefaultedRecord = defaulted(
  record(string(), string()),
  {},
);

const recordA = create(undefined, DefaultedRecord);
const recordB = create(undefined, DefaultedRecord);

recordA.name = 'maddy';

console.log(recordB.name);
// Expect: undefined
// Receive: "maddy"

// OK: object is created through function
// ======================================

const DefaultedRecordFn = defaulted(
  record(string(), string()),
  () => ({}),
);

const recordFnA = create(undefined, DefaultedRecordFn);
const recordFnB = create(undefined, DefaultedRecordFn);

recordFnA.name = 'george';

console.log(recordFnB.name);
// Expect: undefined
// Receive: undefined
```

The current behaviour of a `defaulted(record(...), ...)` is different compared to that of `defaulted(object(...), ...)` and `defaulted(array(...), ...)` which both appear to copy their input data.

yeoffrey:
When using `defaulted` with a `record`, it should copy the default value if its an object. This will fix pass-by-reference bugs with this function.

Closes #1238

### DIFF
--- a/src/structs/coercions.ts
+++ b/src/structs/coercions.ts
@@ -43,7 +43,14 @@ export function defaulted<T, S>(
   } = {}
 ): Struct<T, S> {
   return coerce(struct, unknown(), (x) => {
-    const f = typeof fallback === 'function' ? fallback() : fallback
+    // To avoid a pass-by-reference bug, we'll clone objects when encountered
+    // here, but the for performance avoid cloning primatives and functions
+    const f =
+      typeof fallback === 'function'
+        ? fallback()
+        : typeof fallback === 'object'
+          ? structuredClone(fallback)
+          : fallback
 
     if (x === undefined) {
       return f
--- a/test/defaultedRecord.test.ts
+++ b/test/defaultedRecord.test.ts
@@ -1,0 +1,26 @@
+import { expect, test } from 'vitest'
+import { create, defaulted, record, string } from '../src'
+
+test('Defaulted record value is an empty object', () => {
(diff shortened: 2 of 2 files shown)


arturmuller:
Thanks @yeoffrey for the PR — #1238 was resolved by merging in #1151 which adds coercers to record (and tuple), so I am closing this.

Having had some time to think about it, there are are two things that I realised:

1. The custom class use-case is pretty much irrelevant. Superstruct operates primarily structurally, just like TypeScript. We have `{...value}` and `array.slice()` calls all over the place, so if you're using custom classes you're gonna have a bad time no matter what. This should be added to README somewhere.
2. I think that structs should not touch data that is outside than their responsibility. This means that `structuredClone` is probably not something we want to ever use, since that makes a deep copy which is outside of its responsibility. I think the way superstruct is architected relies on every struct only every making shallow copies and leaving nested logic to subsequent/nested structs. I think this is a big strength, especially when masking and/or using the `type` struct, as it means that if there are parts of the source object that are not defined as a struct, they can be entirely skipped during processing. If the source objects are large, this can end up being quite efficient.

PR #1268, opened 2024-07-27
Replace custom Rollup setup with tsup

arturmuller:


package.json
```diff
@@ -1,13 +1,25 @@
...
   "main": "./dist/index.cjs",
-  "module": "./dist/index.mjs",
+  "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
```
MajorLift, line 16: As discussed, this fix resolves the ["Masquerading as ESM"](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md) issue that affected previous releases, as demonstrated by the ATTW diagnostics results ([before](https://arethetypeswrong.github.io/?p=superstruct%402.0.2) vs. [after](https://arethetypeswrong.github.io/?p=superstruct%402.0.3-1)).

In the current JavaScript ecosystem, as long as `"type": "module"` is specified in package.json, there is no distinction between `.js` vs. `.mjs` and `.d.ts` vs. `.d.mts` files, so this should work without issue. 

That said, all else being equal, I wonder if preferring explicit file extensions (exclusively using `.mjs`, `.d.mts` for ESM) would help future-proof the code against any potential breaking changes in the module system implementations of ECMAScript, Node.js, or TypeScript. Given the complicated history of CJS/ESM-interop this doesn't seem like a remote possibility.

I'm curious whether you see any advantage in using the ambiguous `.js`, `.d.ts` extensions, or if you are indifferent between the two options?

PR #1274, opened 2024-08-20
Run a Codeshift to move away from the custom test runner

yeoffrey:
Related: #1244 

This PR does a few things to the testing suite:
- It removes the old test runner and instead turns each test, which was previously a configuration for a test, into its own vitest `test`.
- it adds a custom matcher to compare `StructError`s, since there is extra logic on comparing `StructError`s in the old test runner and its worth not redefining that check everywhere.

In a follow up I'd like to move all the tests into one file per struct, but that can be done by hand later.


rename.sh
yeoffrey, line 1: Here is the script for renaming the files to end with `.test.ts`

transform.cjs
yeoffrey, line 1: Here is my script for running the Codeshift (warning its actually terrible but whatever)

vitest.config.ts
yeoffrey, line 1: Adding a vitest configuration file for adding the custom matcher

test/matchers.ts
yeoffrey, line 1: File for adding custom matchers, which includes `.toMatchStructError`

The old test runner would check for a failure, and check that the `StructError` gave the right failures but also that the `StructError`'s properties (only the `FILTERED_PROPS` ones) matched the first `Failure` returned by the failures generator on the `StructError`

PR #1300, opened 2026-07-13
fix(assign): preserve input coercion when merging structs

### LINKED ISSUE #791: assign breaks coercion
Wrapping a struct using `defaulted` with `assign` causes coercion to no longer work.

Easy enough test:

```
const struct = defaulted(
  object({
    id: string(),
    enabled: boolean(),
  }),
  {
    enabled: false,
  },
);

console.log(struct.coercer()); // { enabled: false }
console.log(assign(struct).coercer()); // undefined
```

Expected result: coercer still works

xianjianlf2:
## Problem

`assign` rebuilt the merged struct via `object()`/`type()` from the combined schemas, discarding any top-level coercer the input structs carried (e.g. `defaulted`, `trimmed`). As a result, wrapping a struct with `defaulted` and then passing it to `assign` silently dropped the default values during `create`.

## Fix

Compose the inputs' coercers into the merged struct's coercer. The merged struct's own coercer still runs last so masking is applied once against the combined schema, and the inputs run with masking disabled so they don't strip keys owned by the other structs being assigned.

## Testing

Added `test/validation/assign/coerce-defaulted.ts`, which exercises `create` on an `assign`-merged struct wrapped with `defaulted` and asserts the default values are preserved.

Closes #791


xianjianlf2:
Pushed empty commit `cfbec19` to retry CI. The previous Node 16 job failed during `npm install` with `ERR_SOCKET_TIMEOUT`, and the Node 18 job completed tests/lint successfully before GitHub marked the operation canceled.

xianjianlf2:
Fixed the remaining Node 16 CI failure in `971252b` by adding an npm override for `vite@5.4.11`. The failure was in Vitest/Vite startup on Node 16 (`node:crypto.getRandomValues` is unavailable), not in the Superstruct runtime changes.

Local validation with Node 16:
- `npx -p node@16 -c 'node --version; npm --version; npm test && npm run lint'`
- `npm test`: build, type tests, and 226 Vitest tests passed
- `npm run lint`: ESLint and Prettier passed

